PR #4, opened 2023-10-09
ODC-7393: Add type definitions for results api using openapi spec

saumeya:
To test the api, run `npx ts-node src/components/api-type-test.ts` 

Check sample post request input in _**src/components/api-type-test.ts**_


saumeya:
@jerolimov @vikram-raj @lokanandaprabhu

package.json
```diff
@@ -17,7 +17,9 @@
     "test-cypress-headless": "cd integration-tests && node --max-old-space-size=4096 ../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome}",
     "cypress-merge": "mochawesome-merge ./integration-tests/screenshots/cypress_report*.json > ./integration-tests/screenshots/cypress.json",
     "cypress-generate": "marge -o ./integration-tests/screenshots/ -f cypress-report -t 'OpenShift Console Plugin Template Cypress Test Results' -p 'OpenShift Cypress Plugin Template Test Results' --showPassed false --assetsDir ./integration-tests/screenshots/cypress/assets ./integration-tests/screenshots/cypress.json",
-    "cypress-postreport": "yarn cypress-merge && yarn cypress-generate"
+    "cypress-postreport": "yarn cypress-merge && yarn cypress-generate",
+    "generate-types": "openapi-typescript openapi/spec.yml --output src/types/api.ts",
+    "start:dev": "npx nodemon"
```
Lucifergene, line 22: Its not restarting the webpack server properly. 

Better to go with the one used in the [Kubevirt Plugin](https://github.com/kubevirt-ui/kubevirt-plugin/blob/4c03b2d7a68530ef9539d1f52a85a538da935c41/package.json#L176C5-L176C147)

```suggestion
 "dev": "NODE_ENV=development NODE_OPTIONS=--max-old-space-size=8192 rm -rf dist && yarn ts-node ./node_modules/.bin/webpack serve --progress"
```

saumeya: Not needed to run this command - It can be run with this - `npx ts-node src/components/api-type-test.ts`
I am updating the PR to not use express and get the results from consoleFetch and display on the ui, so that there are no additional steps.

src/components/api-type-test.ts
```diff
@@ -0,0 +1,63 @@
+import * as express from "express";
+import { components } from "../types/api";
+
+const list_results: components["schemas"]["Result"][] = [];
```
jerolimov, line 4: This is a nice test of the API. Can you create a file `src/types/index.ts` that exports these subtypes on one level?

The syntax might be a bit different, but something like

```
export type Result = components["schemas"]["Result"]
```

So that you can use this simplified import here:

```suggestion
const results: Result[] = [];
```

src/components/api-type-test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    console.log("parent", parent);
+
+    console.log(list_results);
+  res.json(list_results);
+});
+
+  app.post("/v1alpha2/parents/:parent/results/:result_uid", (req, res) => {
+
+    const result_uid = req.params.result_uid
```
jerolimov, line 21: I understand its only for testing the types, but can you fix that intention anyway? :)

jerolimov (changes requested):
@saumeya Unluckily, this PR needs a rebase. I also added two tiny comments above.

src/components/api-type-test.ts
saumeya: Sure

jerolimov (approved):
Fine for me. We can delete the `api-type-test.ts` as soon as we have integrated the API.

vikram-raj (approved):

PR #12, opened 2023-10-26
ODC-7410: Fix linter issues by running yarn lint plus some manual type fixes

jerolimov:
Depends on https://github.com/openshift/release/pull/44935

Most linter issues were fixed automatically with `yarn lint`.

I fixed the left overs the quick and dirty way by changing any to never and applying some types.

Yes, I also fixed a type issue in the generated API schema. For that, we should extract that file from the linter (by moving it from the src into a gen folder?). But for now, it was more important to me to just get the build green again.


Lucifergene:
Looks good from my side :)

saumeya:
The changes in the generated API schema is double to single quotes or something other than that?

jerolimov:
> The changes in the generated API schema is double to single quotes or something other than that?

There was also one point where it uses `any`, which isn't allowed with the current eslint rules. We have two solutions:

1. Extend `yarn generate-types` to run also eslint for the generated file.
2. Exclude or move the generated file from the linter
3. Configure the generator (don't know that is really needed)

Do you want to take a look?

saumeya:
> > The changes in the generated API schema is double to single quotes or something other than that?
> 
> There was also one point where it uses `any`, which isn't allowed with the current eslint rules. We have at least these options, maybe more:
> 
> 1. Extend `yarn generate-types` to run also eslint for the generated file.
> 2. Exclude or move the generated file from the linter
> 3. Configure the generator (don't know that is really needed)
> 
> Do you want to take a look?

Sure, let me check the options.

vikram-raj (approved):
Looks good. 

/lgtm

PR #13, opened 2023-10-27
ODC-7412: Fix e2e failues by having at least one working cypress test example

jerolimov:
Created one minimal cypress test (based on https://github.com/openshift/console-plugin-template/blob/main/integration-tests/tests/example-page.cy.ts), otherwise `yarn test-cypress-headless` fails with:

```
yarn run v1.22.19
$ cd integration-tests && node --max-old-space-size=4096 ../node_modules/.bin/cypress run --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome}

Can't run because no spec files were found.

We searched for specs matching this glob pattern:

  > /home/christoph/git/openshift-pipelines/console-plugin/integration-tests/tests/**/*.cy.{js,jsx,ts,tsx}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### DIFF
--- a/integration-tests/cypress.config.js
+++ b/integration-tests/cypress.config.js
@@ -11,7 +11,7 @@ module.exports = defineConfig({
     configFile: 'reporter-config.json',
   },
   fixturesFolder: 'fixtures',
-  defaultCommandTimeout: 30000,
+  defaultCommandTimeout: 5000,
   retries: {
     runMode: 1,
     openMode: 0,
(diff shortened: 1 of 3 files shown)


vikram-raj:
The codes look good to me. 

Let's merge https://github.com/openshift-pipelines/console-plugin/pull/12 and then rebase it.

vikram-raj (approved):
/lgtm

PR #17, opened 2023-12-01
ODC-7411: Update i18n and move some parts of readme to contributing

jerolimov:
* Remove some translation prefixes so that we don't create a `console-shared.json` and `public.json` anymore
* I also extracted some contributor info from the README into a new CONTRIBUTING.md so that we can add some screenshots next week to the README. :smile:


vikram-raj (approved):
/lgtm

PR #16, opened 2023-11-10
[WIP] : ODC-7404 Integrate Results API in PipelineRuns card

saumeya:

PR #22, opened 2023-12-07
ODC-7450: Complete Interval Dropdown for API calls

Aneesh-M-Bhat:
PR on top of #16 
Story: [ODC-7450](https://issues.redhat.com/browse/ODC-7450)
Purpose: 
Added intervals for Duration Card, Numbers Chart, StatusCard, TotalRunsCard

### DIFF
--- a/src/components/pipelines-overview/PipelineRunsTotalCard.tsx
+++ b/src/components/pipelines-overview/PipelineRunsTotalCard.tsx
@@ -17,6 +17,7 @@ import { getResultsSummary } from '../utils/summary-api';
 import { DataType } from '../utils/tekton-results';
 import { ALL_NAMESPACES_KEY } from '../../consts';
 import { getDropDownDate } from './dateTime';
+import './PipelineRunsTotalCard.scss';
 
 interface PipelinesRunsDurationProps {
   namespace: string;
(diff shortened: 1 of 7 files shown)


src/components/pipelines-overview/PipelineRunsTotalCard.tsx
```diff
@@ -43,51 +43,44 @@ const PipelinesRunsTotalCard: React.FC<PipelinesRunsDurationProps> = ({
...
       data_type: DataType.PipelineRun,
       filter,
     })
       .then((response) => {
-        setRepoRun(response.summary[0].total);
+        setData(response.summary[0].total);
       })
       .catch((e) => {
-        console.error('Error in getSummary', e);
+        console.error('Unable to Post', e);
```
saumeya, line 57: Nit: revert to Error in getSummary

src/components/pipelines-overview/utils.ts
```diff
@@ -67,3 +67,18 @@ export const LAST_LANGUAGE_LOCAL_STORAGE_KEY = 'bridge/last-language';
 
 export const getLastLanguage = (): string =>
   localStorage.getItem(LAST_LANGUAGE_LOCAL_STORAGE_KEY);
+
+export const useInterval = (
```
saumeya, line 71: Where is the `No refresh` case handled?

saumeya (commented):
Left minor comments, other looks good to me!

src/components/pipelines-overview/utils.ts
Aneesh-M-Bhat: ya so, in the useInterval hook, when user selects no refresh it updates once, then in this custom hook i have put a condition to run setInterval() only if interval!==null, which is the value when you select no refresh.

src/components/pipelines-overview/PipelineRunsTotalCard.tsx
Aneesh-M-Bhat: ya had issue selecting 1 or the other, will revert to error in get summary.

saumeya (approved):
/lgtm

PR #25, opened 2023-12-11
ODC-7455: consmetic changes in Charts and number cards

vikram-raj:
Story: https://issues.redhat.com/browse/ODC-7455

Descriptions: 
1. Aligned the numbers to the right in the Duration and Total runs card
2. Charts axis styles (font and color) should be the same for all charts
3. increase the Chart bar width

image


lokanandaprabhu:
Tested locally.

/lgtm

PR #28, opened 2023-12-13
ODC-7443: Integrate metrics tab for Pipeline and Repository with Summary API

vikram-raj:
image


lokanandaprabhu:
In Average duration card y-axis values and cursor pointing is not proper.

image

src/components/pipelines-metrics/PipelinesMetricsPage.tsx
```diff
@@ -51,6 +40,8 @@ const PipelinesMetricsPage: React.FC = () => {
           domain={{ y: [0, 100] }}
           bordered={false}
           namespace={namespace}
+          parentName={parentName}
```
saumeya, line 43: We need a check when a user directly changes the namespace from the dropdown on the Metrics Tab - it should redirect to the Pipelines tab or show `No metrics found`. Currently gives error on switching. 
The top left corner drop down.

project-drop

saumeya (changes requested):

src/components/pipelines-metrics/PipelinesAverageDuration.tsx
```diff
@@ -115,7 +165,11 @@ const PipelinesAverageDuration: React.FC<PipelinesAverageDurationProps> = ({
                 style={xAxisStyle}
                 tickFormat={xTickFormat}
               />
-              <ChartAxis dependentAxis style={yAxisStyle} />
+              <ChartAxis
+                dependentAxis
+                style={yAxisStyle}
+                tickFormat={(v) => `${v}m`}
```
saumeya, line 171: Can we return in readable format  - it shows decimal values 1.76m instead of 1m 43 s
Screenshot from 2023-12-13 20-33-37

src/components/pipelines-metrics/PipelinesMetricsPage.tsx
saumeya: Error message -
error

saumeya (changes requested):
Just left 2 comments, other than these LGTM

src/components/pipelines-metrics/PipelinesMetricsPage.tsx
vikram-raj: It is a console issue that needs to fix it in the console repository.

PR #30, opened 2023-12-13
OCPBUGS-25334: shows correct data on BarCharts

vikram-raj:
https://drive.google.com/file/d/1XzBy5XWB3lPEI_zz7zxBoOxw-E7VqVUW/view?usp=sharing

image


lokanandaprabhu:
Hi @vikram-raj / @saumeya 

Functionality wise it is working fine but I noticed one thing is we are setting the y-axis max to the max value of the data so you can see the tooltip styling in the attached video. It is trying to compress the tooltip inside the container so the pointer is missing. Maybe we can add it later to consider max value for y-axis as max+buffer of max value of items? 

In Number of PipelineRuns chart if you observe, y-axis max value is 14 and the data max value is 15, that also may be confusing. 

For now I am adding lgtm since we need this functionality in.  

https://github.com/openshift-pipelines/console-plugin/assets/102503482/067f99e3-d844-48e7-92d5-6dafdcb3efe3

/lgtm

PR #35, opened 2023-12-18
ODC-7462: Hide standalone PipelineRuns by req. spec.pipelineRef.name for Pipeline PLRs

jerolimov:
This ignores change to the filter ignores PipelineRuns that are started individually.

Solves also a bug that the PipelineRun name was used as Pipeline name in the Pipeline list page. But the click on the Pipeline shows 404.

**Demo PipelineRun:**

```yaml
apiVersion: tekton.dev/v1
kind: PipelineRun
metadata:
  generateName: independent-pipeline-
spec:
  params:
    - name: NAME
      value: world
  pipelineSpec:
    tasks:
      - name: echo-hello
        taskSpec:
          metadata: {}
          spec: null
          steps:
            - computeResources: {}
              image: ubuntu
              name: echo
              script: |
                #!/usr/bin/env bash
                echo "Hello"
      - name: echo-name
        runAfter:
          - echo-hello
        taskSpec:
          metadata: {}
          spec: null
          steps:
            - computeResources: {}
              image: ubuntu
              name: echo
              script: |
                #!/usr/bin/env bash
                echo "$(params.NAME)"
      - name: echo-bye
        taskSpec:
          metadata: {}
          spec: null
          steps:
            - computeResources: {}
              image: ubuntu
              name: echo
              script: |
                #!/usr/bin/env bash
                sleep 3
                echo "Goodbye!"
  taskRunTemplate:
    serviceAccountName: pipeline
  timeouts:
    pipeline: 1h0m0s
```

**Without this PR:**

image

**With this PR:**

image

### DIFF
--- a/src/components/pipelines-overview/PipelineRunsTotalCard.tsx
+++ b/src/components/pipelines-overview/PipelineRunsTotalCard.tsx
@@ -67,25 +67,25 @@ const PipelinesRunsTotalCard: React.FC<PipelinesRunsDurationProps> = ({
       });
   };
 
-  const filter = `data.metadata.labels.contains("pipelinesascode.tekton.dev/repository")&&data.status.startTime>timestamp("${date}")`;
+  const pipelineFilter = `data.spec.pipelineRef.contains("name") && data.status.startTime>timestamp("${date}")`;
   useInterval(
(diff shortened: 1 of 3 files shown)


saumeya (approved):
LGTM

PR #36, opened 2023-12-18
Add small scripts to connect to the postgres and show the api-log

jerolimov:


scripts/psql-cluster.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e
+
+oc rsh -n openshift-pipelines statefulset/tekton-results-postgres \
```
saumeya, line 5: What does this command do? Do we run queries in it?

jerolimov: It opens `psql` in cluster Pod. A postgresql terminal where you can run SQL queries like

`select * from records;`

You need always close your statement with `;` + enter and you can leave the CLI tool with ctrl+d.

saumeya (approved):
LGTM, thanks!

PR #38, opened 2023-12-18
Create a TypeScript script to create some demo data

jerolimov:
A small script that creates some PipelineRun and save them directly into the Tekton Results DB.

It's written in TypeScript and could be extended based on future needs for e2e tests, etc.

**Idea (why and some thoughts about current issues and opportunities for the future):**

* Instead of creating pure random data, the scripts create some **repeatable** and **predictable** test data. This happens mostly because the scripts calculate the number of PipelineRuns for a day based on that date.
* Weekend shows less PLRs then work days, so that the charts should look a bit more realistic ;)
* Based on the number of PipelineRuns per day, it creates succeeded and failed PipelineRuns for a non-existing Pipeline.
* For a performance test, I would expect that we can just increase these numbers by adding support for some environment variables. (ADDITONAL_PIPELINERUNS_PERDAY) or by adding an alternative script that focuses on that.
* It would be great to create the Pipeline as well and extend it later for PipelineRuns from a Repository.
* The dashboard table shows 404 if the user clicks the Pipeline. But since it is called "demo-data" I hope that's fine for now.
* Instead of creating PipelineRuns that will be executed on the cluster it inserts the data directly into the Postgresql.
* This could result in unexpected errors because the JSON might not match 100% real PipelineRuns. Expect issues when testing this the first time. I used a YAML for the PipelineRun so that missing data could be added easily.

**Usage:**

1. You can open a port-forward to your Tekton Results DB when you are logged in via `oc`:
   ```
   oc login ....
   psql-port-forward.sh
   ```
2. After that you can run the db-* scripts in another terminal. The script uses `oc` to get the database credentials from a cluster and connects then against `localhost:5432`.

   ```
   cd scripts
   
   ./db-create-demo-data.sh
   
   # should print some info about created DB entries
   
   Create 380 / 380 PipelineRuns between Mon, Nov 20, 2023 and Mon, Dec 18, 2023
   Done in 8.39s.
   
   
   ./db-show-stats.sh                              
   
   Number of PipelineRuns since 2023-12-11T17:18:21.483Z
   - All PipelineRuns: 0
   - Started from a Pipeline: 0
   - Started from a Repository: 0
   - Standalone PipelineRuns: 172
   Done in 2.19s.
   ```

The result should look like this:

image


saumeya (commented):
Works well for me @jerolimov ,  I can see the data in my console dev env. Thanks, this will help in development.
We can add more improvements as you suggested and make the scripts more interactive - 
Like specifying the number of days for showing stats or the start date for creating pipeline runs.

saumeya (approved):
LGTM

PR #43, opened 2023-12-29
fix: For 1st plr show - for last runtime until it is completed

lokanandaprabhu:
Description:

If it is 1st PLR for pipeline and repository, last run time will be shown as '-'. From API last run time is passed as completed timestamp, so show '-' till it gets completed


saumeya (approved):
/lgtm

PR #44, opened 2024-01-03
ODC-7466: Fix design issue as per the UX design

lokanandaprabhu:
**Story**: https://issues.redhat.com/browse/ODC-7466

**Screenshots**:

--**Before**---

Screenshot 2024-01-03 at 1 12 17 PM

Screenshot 2024-01-03 at 1 23 01 PM

--**After**---

Screenshot 2024-01-03 at 1 00 23 PM

Screenshot 2024-01-03 at 1 22 27 PM

### DIFF
--- a/src/components/pipelines-metrics/PipelinesAverageDuration.tsx
+++ b/src/components/pipelines-metrics/PipelinesAverageDuration.tsx
@@ -152,7 +152,7 @@ const PipelinesAverageDuration: React.FC<PipelinesAverageDurationProps> = ({
               domain={domainValue}
               domainPadding={{ x: [30, 25] }}
               height={150}
-              width={400}
+              width={500}
               padding={{
                 top: 20,
                 bottom: 40,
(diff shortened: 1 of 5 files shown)


vikram-raj (commented):
Loader should be align with the title. Needs to add left padding
image

Can we make the Number of PipelineRuns chart to cover the 100% width of the card when it move to next line on small screen.
image

Column sort indicator arrow color should be light check the other tables in console.
image

See this 
image

lokanandaprabhu:
> Loader should be align with the title. Needs to add left padding
Resolved it

> Can we make the Number of PipelineRuns chart to cover the 100% width of the card when it move to next line on small screen.
I looked into it. If I don't specify the width, svg width is getting set to 450 and this width is not getting changed even if parent div width increases and there is no separate class defined for svg to manually set the width based on screen size. I will take a look at it again. Please guide here if you have any idea.

> Column sort indicator arrow color should be light check the other tables in console.
Class and the values are proper but PF variable is not getting added. May be if we upgrade the PF this issue will be resolved.

jerolimov (approved):
/lgtm

PR #60, opened 2024-02-05
ODC-7515: add locale files for ja, ko, and zh

vikram-raj:


jerolimov (approved):
Great first step :+1:

/lgtm

vikram-raj:
/retitle ODC-7515: add locale files for ja, ko, and zh

PR #62, opened 2024-02-05
SRVKP-4088: sync docker login and buildx gh action

jkandasa:
Fixes the following issue by syncing the docker login and buildx GH action versions,
```
docker/setup-buildx-action@v1 is not allowed to be used in openshift-pipelines/console-plugin. Actions in this workflow must be: within a repository owned by openshift-pipelines, created by GitHub, or matching the following: docker/login-action@*.
```

### DIFF
--- a/.github/workflows/publish_container_image.yaml
+++ b/.github/workflows/publish_container_image.yaml
@@ -17,7 +17,7 @@ jobs:
         with:
           node-version: '16'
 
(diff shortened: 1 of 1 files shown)


jkandasa:
cc @vdemeester

vdemeester:
/approve
/lgtm

It's not gonna fix the problem, but it updates the version so 👍🏼 . I did allow the action on the repository for it to work now.

PR #65, opened 2024-02-11
SRVKP-4088: [main] tune multiarch container image build time

jkandasa:
* performing `yarn install` on multiarch with `docker buildx` takes a lot of time (approx `2h30m`)
 to avoid this, `yarn install` and `yarn build` should be executed outside of docker (on a host platform)
 and only copy the `dist` to different platform/architecture runtime containers
**WARNING:** with this approach, if we use platform native package may lead to a malfunction
 however in this project, the code will be executed on a browser and assuming there is no platform specific implementations, hence assuming no impact on this approach

* and builds the container image for `release-v*` branches

### DIFF
--- a/.github/workflows/publish_container_image.yaml
+++ b/.github/workflows/publish_container_image.yaml
@@ -1,12 +1,14 @@
 name: publish container images
 on:
   push:
-    branches: [main]
+    branches:
+      - main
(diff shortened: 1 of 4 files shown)


jkandasa:
cc @vdemeester

vdemeester (approved):
/lgtm

PR #74, opened 2024-02-21
OCPBUGS-29816: Switch to service from external result route endpoint

vikram-raj:


khrm (commented):
I think we need to do some change in other pages also. I saw overview page working fine. But pipeline's pipelinerun's page fail

PR #76, opened 2024-03-01
OCPBUGS-30160: Non-admin users are not able to see overview page

lokanandaprabhu:
**Issue**: https://issues.redhat.com/browse/OCPBUGS-30160

**Description**: As developer user, cluster version is not accessible from config.openshift.io resource. So added to fetch from window.SERVER_FLAGS

### DIFF
--- a/src/types/openshift.ts
+++ b/src/types/openshift.ts
@@ -10,35 +10,3 @@ export type K8sResourceKind = K8sResourceCommon & {
   status?: { [key: string]: any };
   data?: { [key: string]: any };
 };
-
-export type UpdateHistory = {
-  state: 'Completed' | 'Partial';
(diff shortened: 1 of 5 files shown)


src/types/openshift.ts
```diff
@@ -26,19 +26,3 @@ export type Release = {
   url?: string;
   channels?: string[];
 };
```
vikram-raj, line 28: We can remove `UpdateHistory` and `Release` type definitions.

lokanandaprabhu: Updated the PR

vikram-raj (commented):
/lgtm

khrm (commented):
/cherry-pick release-v1.14.x

PR #89, opened 2024-04-09
ODC-7559: Improve PLR list page performance

lokanandaprabhu:
**Story:**
https://issues.redhat.com/browse/ODC-7559


jerolimov:
Instead of fetching all TaskRuns and merging 10000 TR with 100 PLR, it now fetches now only the TRs that are NOT successed. @lokanandaprabhu will continue to improve this so that we also don't fetch the TR for FAILED.

fyi @vikram-raj https://redhat-internal.slack.com/archives/CHG0KRB7G/p1713778383616389

jerolimov (approved):
/lgtm

PR #97, opened 2024-05-02
ODC-7558: Add all Projects list page

vikram-raj:


lokanandaprabhu:
https://github.com/openshift-pipelines/console-plugin/assets/102503482/91b2c7db-41cb-4ebe-8a9b-407ccba9e591

@vikram-raj , can we add some spacing so that it look similar to other pages? and we are not considering `Create a Project` button besides "Select a Project" now?

lokanandaprabhu (approved):

lokanandaprabhu:
Tested locally. Working fine.

/lgtm

PR #112, opened 2024-05-20
OCPBUGS-34574: TaskRuns should not be fetched for Failed PLR's

lokanandaprabhu:
Static plugin PR: https://github.com/openshift/console/pull/13806


vikram-raj (commented):
verified

/lgtm

PR #119, opened 2024-05-28
ODC-7584: Add filter to fetch from cluster and tekton results in TaskRun list pages

lokanandaprabhu:
**Story:**
https://issues.redhat.com/browse/ODC-7584

**Video:**
https://drive.google.com/file/d/12M-tFBk0L2fMXP7B9272J2Y8awno4I6n/view?usp=drive_link

**Description:**
1. Added another filter group for the TaskRun list pages which will have 2 options, one is to fetch live data from cluster and another one is to fetch archived data from tekton results. 

2. If Tekton Results is not installed, then the data source filter group will not be visible

3. By default, cluster data source will be selected

4. Changes are done in below pages
     a) TaskRun list page
     b) TaskRun list page tab in PipelineRun details page


vikram-raj:
cc: @koustavsaha

vikram-raj (commented):
/lgtm

PR #123, opened 2024-06-11
ODC-7600: Add Timeout field in the start pipeline modal

vikram-raj:
Add a Timeout field in the start pipeline modal

https://github.com/openshift/console/assets/2561818/336610c9-ec87-44d6-8c18-c31a8e12ab94


lokanandaprabhu (changes requested):
If I don't enter any value in timeout I am getting error message

Screenshot 2024-06-12 at 12 10 30 PM

src/components/start-pipeline/utils.ts
```diff
@@ -331,8 +331,8 @@ export const getPipelineRunFromForm = (
   annotations?: { [key: string]: string },
   options?: { generateName: boolean },
 ) => {
-  const { parameters, workspaces } = formValues;
-
+  const { parameters, workspaces, timeouts } = formValues;
+  console.log(timeouts, '####');
```
lokanandaprabhu, line 335: Can you please remove this log

lokanandaprabhu:
Due to new form group, the space is more above Timeouts section, can we add this as part of Parameter section or remove the additional space? 

Screenshot 2024-06-12 at 12 32 00 PM

vikram-raj:
> Due to new form group, the space is more above Timeouts section, can we add this as part of Parameter section or remove the additional space?

I added it in a different section as it is not a part of the parameters. Timeout is a different section.

lokanandaprabhu (approved):

PR #127, opened 2024-07-01
ODC-7619: Using tekton results utils from aonic library

lokanandaprabhu:
**Stroy:** 

https://issues.redhat.com/browse/ODC-7619

**NOTE:**

AONIC-UI is using React 18 and console does not allow dynamic plugin to upgrade to React 18 since console is running on React 17. So this PR is tested with downgrading AONIC-UI to react 17


lokanandaprabhu:
/close 

We can't merge this PR since aonic-ui is using React 18 and console and console-plugin repo is using React 17. 

When console and console-plugin updated to React 18, create PR that time.

PR #129, opened 2024-07-03
OCPBUGS-33521: Update sorting in pipeline, pipelineRun and TaskRun list pages

lokanandaprabhu:
**Issue:**
https://issues.redhat.com/browse/OCPBUGS-33521

**Description:**
Pipeline list page - by default sorting will happen on `Last run time` column
PipelineRun list page - by default sorting will happen on `Started` column
TaskRun list page - by default sorting will happen on `Started` column

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,8 @@
   },
   "devDependencies": {
     "@cypress/webpack-preprocessor": "^5.15.5",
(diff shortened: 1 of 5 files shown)


vikram-raj (commented):
/lgtm

PR #160, opened 2024-10-17
SRVKP-6637: Remove All options from project dropdown on overview page for user have no access to all projects

lokanandaprabhu:
**Fixes:** 
https://issues.redhat.com/browse/SRVKP-6637

----Admin user---

https://drive.google.com/file/d/1FKw1trTy3Gu9Tv0EhRr6_1S_ONhtbQK-/view?usp=drive_link

----Non-admin user who don't have access to All namespace---

https://drive.google.com/file/d/1uh-uwrxNtGNvg0Pd4z-in2vQJJf-LInh/view?usp=drive_link

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -144,7 +144,10 @@
     "type": "console.page/route",
     "properties": {
       "exact": true,
-      "path": ["/pipelines-overview/ns/:ns"],
+      "path": [
+        "/pipelines-overview/ns/:ns",
+        "/pipelines-overview/all-namespaces"
+      ],
       "component": {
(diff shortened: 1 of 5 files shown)


vikram-raj (commented):
I am seeing 404 not found error page for /pipelines-overview/all-namespaces in case of non-admin. and it feels like what to do next. Instead 404 can we have a list of all the projects same as we have in dev perspective for non admin user? if they visit all namespace URL.

lokanandaprabhu:
Updated the PR and added Project selection page for non-admin user if the namespace is all.

vikram-raj (commented):
/lgtm

PR #171, opened 2024-10-25
Add dockerfile for console plugin

savitaashture:


openshift/dockerfiles/console-plugin.Dockerfile
```diff
@@ -0,0 +1,32 @@
+ARG BUILDER=registry-proxy.engineering.redhat.com/rh-osbs/rhacm2-yarn-builder:rhel_8_nodejs_18_yarn_builder
```
vdemeester, line 1: I am not sure this will work (not behind the VPN).

vdemeester (commented):
Could we re-use the `Dockerfile` on the root of this git repo ?

savitaashture:
> Could we re-use the `Dockerfile` on the root of this git repo ?

Ya its almost same 

but we add some labels and all

and also the component which is coming from hack repo alwasy points to openshift folder so

openshift/dockerfiles/console-plugin.Dockerfile
savitaashture: ya changed to registry.access.redhat.com/ubi8/nodejs-18:latest

vdemeester:
The `Dockerfile` is also referenced by the `prow` build, and this is probably why `ci/prow/images` fails 😅

savitaashture:
> The `Dockerfile` is also referenced by the `prow` build, and this is probably why `ci/prow/images` fails 😅

:thinking: 
how to fix this

vdemeester:
@savitaashture symbolic link or duplicate ?

savitaashture:
> @savitaashture symbolic link or duplicate ?

I mistakenly deleted original Dockerfile from root directory
added back now lets see

vdemeester:
@savitaashture that works 😉 
/approve
/lgtm

PR #177, opened 2024-11-05
Add rpms as its required for prefetch dependancy

savitaashture:


savitaashture:
Hi @vdemeester I had to have local dependency in order to fetch and build with yarn and Konflux won't have support for yarn yet like npm, rpm etc...
Jira : https://issues.redhat.com/browse/STONEBLD-1753 (Support of pre-fetch for Yarn  )

So I followed https://classic.yarnpkg.com/blog/2016/11/24/offline-mirror/ to have local copy so that yarn can download offline way

vdemeester:
@savitaashture 🆗

PR #200, opened 2024-12-04
Add release plan for nightly release

savitaashture:


pramodbindal:
@savitaashture 

I think release plan is being generated from hack now so do we need this PR ?

PR #204, opened 2024-12-10
SRVKP-6898: Replace font awesome with custom css in dynamic plugin

lokanandaprabhu:
**Fixes:**
https://issues.redhat.com/browse/SRVKP-6898

### DIFF
--- a/src/components/styles/common.scss
+++ b/src/components/styles/common.scss
@@ -1,0 +1,18 @@
+@keyframes co-spin {
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
+
+.co-spin {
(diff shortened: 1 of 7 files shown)


vikram-raj (commented):
/lgtm

src/components/styles/common.scss
```diff
@@ -0,0 +1,18 @@
+@keyframes co-spin {
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
+
+.co-spin {
```
spadgett, line 10: Don't redefine the same CSS as console provides. I'd avoid adding any CSS with a `co-` prefix and instead use your plugin name as a prefix.

PR #206, opened 2024-12-10
SRVKP-6898: Replace font awesome with custom css in dynamic plugin #204

lokanandaprabhu:
**Description:**
Changed co-spin classname to co-spin-pipelines to aviod conflict with console class name

### DIFF
--- a/src/components/styles/common.scss
+++ b/src/components/styles/common.scss
@@ -1,4 +1,4 @@
-@keyframes co-spin {
+@keyframes pipelines-console-plugin__spin {
   0% {
     transform: rotate(0deg);
   }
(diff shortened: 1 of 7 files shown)


src/components/styles/common.scss
```diff
@@ -7,12 +7,10 @@
   }
 }
 
-.co-spin {
-  animation: co-spin 2s infinite linear;
-}
+.co-spin-pipelines {
```
spadgett, line 10: I'd avoid the `co-` prefix altogether in your plugin. Then you never risk conflicting with console classes that always use the prefix. Something like

```suggestion
.pipelines-console-plugin__spin {
```

would be better imo. See the recommendations here:

https://docs.openshift.com/container-platform/4.17/web_console/dynamic-plugin/overview-dynamic-plugin.html#general-plug-in-guidelines

lokanandaprabhu: Updated the PR. PTAL.

spadgett (approved):
/lgtm

Thanks 👍

vikram-raj (commented):
/lgtm

PR #309, opened 2025-03-11
NO-JIRA: Update install result script

vikram-raj:


scripts/install-results.sh
```diff
@@ -27,19 +10,8 @@ metadata:
...
-  logs_path: /logs
-  logs_type: File
-  logs_buffer_size: 32768
-  auth_disable: true
-  tls_hostname_override: tekton-results-api-service.openshift-pipelines.svc.cluster.local
-  db_enable_auto_migration: true
-  server_port: 8080
-  prometheus_port: 9090
+  loki_stack_name: logging-loki
+  loki_stack_namespace: openshift-logging
```
lokanandaprabhu, line 14: Can you please add also under spec ?

`auth_disable: true`

vikram-raj: added.

lokanandaprabhu (approved):
/lgtm

PR #331, opened 2025-03-24
feat: Implement support for Task `spec.displayName`

aThorp96:
Implements https://issues.redhat.com/browse/SRVKP-5501

### DIFF
--- a/src/components/pipelines-tasks/TasksRow.tsx
+++ b/src/components/pipelines-tasks/TasksRow.tsx
@@ -18,6 +18,8 @@ import {
 } from '@patternfly/react-core';
 import EllipsisVIcon from '@patternfly/react-icons/dist/esm/icons/ellipsis-v-icon';
 import { K8sCommonKebabMenu } from '../utils/k8s-common-kebab-menu';
+import { TaskKind } from 'src/types';
+import { getTaskName } from '../utils/pipeline-augment';
 
 type TasksKebabProps = {
   obj: K8sResourceCommon;
@@ -62,7 +64,7 @@ const TaskKebab: React.FC<TasksKebabProps> = ({ obj }) => {
   );
 };
 
-const TaskRow: React.FC<RowProps<K8sResourceCommon>> = ({
+const TaskRow: React.FC<RowProps<TaskKind>> = ({
   activeColumnIDs,
   obj,
 }) => {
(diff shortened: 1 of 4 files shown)


aThorp96:
@vikram-raj Do you mind reviewing when you are able?

src/components/pipelines-tasks/TasksRow.tsx
```diff
@@ -73,7 +75,7 @@ const TaskRow: React.FC<RowProps<K8sResourceCommon>> = ({
       <TableData activeColumnIDs={activeColumnIDs} id="name">
         <ResourceLink
           kind={getReferenceForModel(TaskModel)}
-          name={obj.metadata.name}
+          name={getTaskName(obj)}
```
vikram-raj, line 78: It will break the details page link. Should make use of `displayName` props
```suggestion
          name={obj.metadata.name}
          displayName={getTaskName(obj)}
```

IMO we should also update the name on the details page. It might confuse the user to see the difference in name in list page and details page.

vikram-raj: https://github.com/user-attachments/assets/d9aed72b-3107-4d77-a014-38dd288a7e4a

vikram-raj (changes requested):

vikram-raj (commented):
The list page uses `spec.displayName`, whereas the details page uses `metadata.name` as a user it looks broken because of difference in name. As this is a feature, can you confirm it with PM/UX?

lokanandaprabhu (commented):
Screenshot 2025-05-07 at 11 47 35 AM

In TaskRun list page, it still shows name instead of displayName. Is this fine?

vikram-raj:
@lokanandaprabhu IMO, we should show displayName here if it is available in TR.  wdyt @aThorp96 ?

aThorp96:
@vikram-raj @lokanandaprabhu in that view the `Task` column's display name is not actually related to the Task object, but the name of the Task in the Pipeline. If the task is not a 

This column could be changed to fallback on the Task's Display Name if the TaskRun is not associated with a Pipeline/PipelineRun, but I worry that could introduce further confusion since the column would then refer to two different fields, depending on the TaskRun.

vikram-raj:
@aThorp96 @lokanandaprabhu, in my opinion, let's go with Task `metadata.name` in TR list page. and lets see if we get any feedback on this from the users.

vikram-raj (commented):
/lgtm

PR #407, opened 2025-04-24
ODC-7789: Remove react-router 5.x

vikram-raj:
**Descriptions:** Remove the react-router 5.x dependency from Pipeline console plugin as the console team is planning to drop React Router 5 support for dynamic plugins

Components that were replaced:
- Replace { history } >> const params = useNavigate();
- Replace { location } >> const location = useLocation();
- Now import `Link` from `react-router-dom-v5-compat`
- Remove RouterComponentProps instead of it use hooks `useParams` to get the params.


lokanandaprabhu (approved):
/lgtm

PR #498, opened 2025-05-29
chore: yarn update, use root for build

waveywaves:


pramodbindal:
@waveywaves 
Removing yarn.lock will make the  konflux builds to fail . Please update the hack repo as well so it does not use the yarn

PR #499, opened 2025-05-29
ODC-7795: Move Setup GitHub App page to dynamic plugin

lokanandaprabhu:
**Description:**

Migrated Setup GitHub App page from static plugin to console dynamic plugin

**Tests:**

image


src/components/pac/__tests__/PacForm.spec.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import { shallow, ShallowWrapper } from 'enzyme';
```
vikram-raj, line 2: As we are migrating from ezyme test to React testing library, Its better to migrate the enzyme unit tests to RTL while moving to console-plugin.

lokanandaprabhu:
Rebased the PR.

PR #510, opened 2025-06-04
Tests to install pipeline operator version 1.17

sanketpathak:
Due to the certificate issue with 4.18, tests should install pipeline version 1.17 until the issue is resolved


vikram-raj (commented):
/lgtm

PR #519, opened 2025-06-05
ODC-7788: Contribute Pipeline related extensions from Pipelines console plugin

lokanandaprabhu:
**Story:**

https://issues.redhat.com/browse/ODC-7788

**Description:**

Moved below extensions to console dynamic plugin

dev-console.add/action - already moved in first phase
console.yaml-template
console.flag/model
console.page/route - already moved in first phase

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -19,6 +19,105 @@
         "kind": "Pipeline"
       },
       "flag": "OPENSHIFT_PIPELINE"
+    },
+    "flags": {
+      "required": ["HIDE_STATIC_PIPELINE_PLUGIN_CONSOLE_FLAG_MODEL_EXTENSION"]
+    }
+  },
+  {
+    "type": "console.flag/model",
+    "properties": {
+      "model": {
+        "group": "pipelinesascode.tekton.dev",
+        "version": "v1alpha1",
+        "kind": "Repository"
+      },
+      "flag": "OPENSHIFT_PIPELINE_AS_CODE"
+    },
+    "flags": {
+      "required": ["HIDE_STATIC_PIPELINE_PLUGIN_CONSOLE_FLAG_MODEL_EXTENSION"]
+    }
+  },
+  {
+    "type": "console.flag/model",
+    "properties": {
+      "model": {
+        "group": "tekton.dev",
+        "version": "v1beta1",
+        "kind": "Pipeline"
+      },
+      "flag": "OPENSHIFT_PIPELINE_V1BETA1"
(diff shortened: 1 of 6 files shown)


console-extensions.json
```diff
@@ -19,6 +19,136 @@
...
+      "required": ["HIDE_STATIC_PIPELINE_PLUGIN_CONSOLE_FLAG_MODEL_EXTENSION"]
+    }
+  },
+  {
+    "type": "console.flag/model",
+    "properties": {
+      "model": {
+        "group": "tekton.dev",
+        "version": "v1alpha1",
+        "kind": "Condition"
```
vikram-raj, line 33: No need to migrate conditions as it is deprecated.

console-extensions.json
```diff
@@ -19,6 +19,105 @@
...
+  },
+  {
+    "type": "console.flag/model",
+    "properties": {
+      "model": {
+        "group": "tekton.dev",
+        "version": "v1",
+        "kind": "Pipeline"
+      },
+      "flag": "OPENSHIFT_PIPELINE_V1BETA1"
```
vikram-raj, line 49:
```suggestion
      "flag": "OPENSHIFT_PIPELINE"
```

src/components/templates/pipelines.ts
```diff
@@ -0,0 +1,76 @@
...
+  resources:
+    - name: app-git
+      type: git
+    - name: app-image
+      type: image
+  tasks:
+    - name: build-app
+      taskRef:
+        name: s2i-java-11
+        kind: ClusterTask
```
vikram-raj, line 19: We should update this Pipeline template, as it uses ClusterTask, which is already deprecated and removed.

src/components/templates/pipelines.ts
```diff
@@ -0,0 +1,76 @@
...
+  steps:
+  - image: registry.redhat.io/ubi7/ubi-minimal
+    command:
+    - /bin/bash
+    - '-c'
+    - echo
+    - $(inputs.params.appName)
+`;
+
+export const newTaskRunTemplate = `
```
vikram-raj, line 61: I do not see this template when creating a new TaskRun and same happening with new Pipeline.

lokanandaprabhu: For TaskRun I resolved, for Pipeline due to /builder page, this default template is not adding, If I remove /builder from URL i can see the template.

vikram-raj (approved):
/lgtm

PR #555, opened 2025-06-20
chore: update with working FIPs enabled dockerfile

waveywaves:


waveywaves:
build is failing because of prefetch dependencies not being set as they have been set in the release-v1.19.0 branch

PR #565, opened 2025-06-24
SRVKP-5546: Pipelines builder shows only some tasks from hubs

lokanandaprabhu:
**Fixes:**

https://issues.redhat.com/browse/SRVKP-5546

**Description:**

Call to ArtifactHub was happening only once and if user search any task and if that task was not returned in first API call, the task was not listed in quick search. With this fix, if user searched something, API call will be made to ArtifactHub api and fetch the task

**Screen recording:**

ODC-artifactHub-search-issue-handled

### DIFF
--- a/src/components/catalog/apis/artifactHub.ts
+++ b/src/components/catalog/apis/artifactHub.ts
@@ -182,3 +182,21 @@ export const updateArtifactHubTask = async (
     throw err;
   }
 };
+
+export const fetchArtifactHubTasks = async (
+  query: string,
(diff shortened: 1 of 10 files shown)


vikram-raj:
I don't think I was clear. I meant we should also include display_name instead of name in the list.

lokanandaprabhu:
@vikram-raj Updated the PR to handle the issue. PTAL.

/retest

vikram-raj (approved):
/lgtm

PR #633, opened 2025-08-26
feat: Enable 'All' namespaces view for non-admin users

mathur07:
- Show 'All' option in namespace dropdown for all users with accessible namespaces
- Remove admin-only restriction from pipelines overview pages
- Non-admins see pipeline data scoped to their accessible namespaces
- Maintains security through existing RBAC permissions

Fixes: [SRVKP-6766](https://issues.redhat.com//browse/SRVKP-6766)

### DIFF
--- a/src/components/pipelines-overview/NamespaceDropdown.tsx
+++ b/src/components/pipelines-overview/NamespaceDropdown.tsx
@@ -58,11 +58,15 @@ const NameSpaceDropdown: React.FC<NameSpaceDropdownProps> = ({
       items.push({ title: selected, key: selected }); // Add current namespace if it isn't included
     }
     items.sort((a, b) => alphanumericCompare(a.title, b.title));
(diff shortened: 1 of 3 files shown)


mathur07:
Hey @vikram-raj @lokanandaprabhu @pratap0007 and @aThorp96, I have a this [SRVKP-6766](https://issues.redhat.com//browse/SRVKP-6766) that needs refinement. I was looking through the code and saw that it may be possible to implement it through the changes in this PR. I would need your advice to guide me if this change will be the correct fix to address "Show 'All' option in namespace dropdown for non-admin users with respective accessible namespaces". If not, then what is the correct approach that should be followed to achieve the desired states. Based on that we will be able to break the story into tasks and epics. TY!

anwesha-palit-redhat:
Closing this as there is no response

PR #678, opened 2025-10-13
SRVKP-8998: fix Pipeline Builder UI shows stale parameter data for tasks with same name across namespaces

anwesha-palit-redhat:
## **Description of the Problem**

The customer experiences confusion and incorrect task configuration when editing pipelines through the **Pipeline Builder UI**.
Because the side panel displays **stale or incorrect parameter data** for tasks that share the same name across namespaces, users may unknowingly save or deploy pipelines with invalid configurations.
This leads to **pipeline failures** and **loss of confidence** in the UI’s reliability.

---

## **Workaround**

A **partial workaround** exists:
Users can manually verify and edit the **Pipeline YAML** to ensure it references the correct task and parameters from the intended namespace.
However, this requires advanced Tekton knowledge and defeats the purpose of the visual builder — making it unsuitable for most users.

---

## **Versions**

* **OCP Version:** 4.16
* **Pipelines Operator Version:** 1.17, 1.18

---

## **Steps to Reproduce**

1. Create a task with a name identical to an existing task in the `openshift-pipelines` namespace.
2. Using the **Pipeline Builder UI**, create a new pipeline and add the task from the `openshift-pipelines` namespace.
3. Observe that the **Pipeline Builder side panel** shows parameters from the task in the current namespace instead of the selected one.
4. Check the **YAML editor** — it correctly references the task from the cluster resolver and displays the accurate parameter data.

---

## **Reproducibility**

* **Yes**, reproducible **Always**

---

## **Customer Impact**

* **Customer Name:** Jeffrey Luckett
* **Revenue Impact:** Medium — misconfigured pipelines can lead to failed runs and CI/CD delays, increasing support overhead and reducing user trust in the console plugin UI.

---

## **Root Cause**

In the previous implementation, installation of **Artifact Hub tasks** failed because tasks with the same names already existed in the cluster resolver or namespace.
As a result:

* The data displayed in the **Pipeline Builder form** became stale.
* When tasks were removed from the Pipeline Builder, the actual cluster task was **not uninstalled**.
* Upon reinstalling a task with the same name, the old (stale) data persisted.

---

## **Fix Implemented**

* Added a **check for existing tasks** with the same name before installation.
* If a conflict is found, a **safe name** is generated and used for creating a new task.
* During task removal, the system:

  * Checks if a task with the same name exists.
  * If yes, **uninstalls** it to maintain data hygiene.
  * If not, it updates the **React state** accordingly.

This ensures consistent task data and prevents stale or incorrect parameter references.

---

## **Screen Recordings**

(Since GitHub upload size limits apply, videos are uploaded to Google Drive)

* 🎥 [[Before Fix – Showcasing Issue](https://drive.google.com/drive/folders/1THodXuVY6VyDOf_xiMcc8uOKBh7fP-LC)]
* 🎥 [[After Fix – Validation](https://drive.google.com/drive/folders/1THodXuVY6VyDOf_xiMcc8uOKBh7fP-LC)]

### DIFF
--- a/src/components/pipeline-builder/PipelineBuilderForm.tsx
+++ b/src/components/pipeline-builder/PipelineBuilderForm.tsx
@@ -105,7 +105,6 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
   const updateTasks = (changes: CleanupResults): void => {
     const { tasks, listTasks, finallyTasks, finallyListTasks, loadingTasks } =
       changes;
-
     setFieldValue('formData', {
       ...formData,
       tasks,
@@ -179,6 +178,20 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
     }
   }, [selectedTask]);
 
+  const handleRemoveTask = React.useCallback(
+    async (taskName: string) => {
+      setSelectedTask(null);
+      updateTasks(
+        applyChange(
+          taskGroup,
+          { type: UpdateOperationType.REMOVE_TASK, data: { taskName } },
+          namespace,
+        ),
+      );
+    },
+    [namespace, taskGroup, updateTasks],
+  );
+
   return (
     <Drawer isExpanded={!!selectedTask} position="right">
       <DrawerContent
@@ -206,19 +219,7 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
               onRemoveTask={(taskName: string) => {
                 launchModal(RemoveTaskModal, {
                   taskName,
-                  onRemove: () => {
-                    setSelectedTask(null);
-                    updateTasks(
-                      applyChange(
-                        taskGroup,
-                        {
-                          type: UpdateOperationType.REMOVE_TASK,
-                          data: { taskName },
-                        },
-                        namespace,
-                      ),
-                    );
-                  },
+                  onRemove: () => handleRemoveTask(taskName),
                 });
               }}
               selectedData={selectedTask}
--- a/src/components/task-quicksearch/PipelineQuickSearch.tsx
+++ b/src/components/task-quicksearch/PipelineQuickSearch.tsx
@@ -20,6 +20,7 @@ import {
   TaskProviders,
   updateTask,
 } from './pipeline-quicksearch-utils';
+import { safeName } from '../pipeline-builder/utils';
 import PipelineQuickSearchDetails from './PipelineQuickSearchDetails';
 import { CatalogServiceProvider } from '../catalog/service';
 import { CatalogService } from '../catalog/types';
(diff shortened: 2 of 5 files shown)


src/components/task-quicksearch/PipelineQuickSearch.tsx
```diff
@@ -64,6 +65,24 @@ const Contents: React.FC<
   useLoadingTaskCleanup(onUpdateTasks, taskGroup);
   useCleanupOnFailure(failedTasks, onUpdateTasks, taskGroup);
 
+  // Get all existing task names from taskGroup and installed tasks
+  const getExistingTaskNames = (): string[] => {
```
jhadvig: the current implementation does not count with duplicated names. I would suggest something like:

```ts
  const taskNames = new Set<string>();

  // Add all tasks currently in the builder
  [
    ...taskGroup.tasks,
    ...taskGroup.finallyTasks,
    ...taskGroup.listTasks,
    ...taskGroup.loadingTasks,
    ...taskGroup.finallyListTasks,
  ].forEach((t) => {
    if (t?.name) taskNames.add(t.name);
  });

  // Add installed catalog items (avoid duplicates)
  catalogService.items.forEach((catalogItem) => {
    const name = catalogItem.data?.metadata?.name;
    if (name) taskNames.add(name);
  });

  return Array.from(taskNames);
};
```

src/components/task-quicksearch/PipelineQuickSearch.tsx
```diff
@@ -143,19 +180,46 @@ const Contents: React.FC<
...
-            );
-            createArtifactHubTask(
-              selectedVersionUrl,
-              namespace,
-              selectedVersion,
-              isDevConsoleProxyAvailable,
-            ).catch(() =>
-              setFailedTasks([...failedTasks, item.data.task.name]),
-            );
+            // Checking if task with same name already exists, if yes then create with a different name to avoid conflict
```
jhadvig: nit: These two blocks are pretty much the same, just using a different model. Lets rather create a callback which will take the model and its parameters

src/components/pipeline-builder/PipelineBuilderForm.tsx
```diff
@@ -206,8 +209,28 @@ const PipelineBuilderForm: React.FC<PipelineBuilderFormProps> = (props) => {
...
+                      });
+                      taskExists = true;
+                    } catch (error) {
+                      console.warn(
+                        `Error fetching Task as task does not exist ${taskName}, so safely deleting from PipelineBuilder`,
+                        error,
+                      );
+                    }
+                    if (taskExists) {
+                      await k8sDelete({
```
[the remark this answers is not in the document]
anwesha-palit-redhat: That’s a good point — Using installedFrom limits cleanup to builder-installed tasks, but users may also install or update tasks via CLI / YAML. In this case, they were installing tasks using CLI and then adding them via the builder.

Therefore, I implemented this as I felt a more general approach might be safer long term — one that ensures consistent cleanup while still protecting user-managed resources.

src/components/task-quicksearch/PipelineQuickSearch.tsx
anwesha-palit-redhat: Done, thanks!

PR #709, opened 2025-10-23
SRVKP-8976: fix correctly selects nav tab via log URL

anwesha-palit-redhat:
###  Description

When navigating directly to a **finally task** using URL parameters (`?taskName=<task>`), the log viewer does **not auto-select** the corresponding task tab.

---

###   Root Cause

The previous matching logic:

```ts
const activeTaskRun = activeTask
  ? taskRuns.find((taskRun) => taskRun.metadata.name.includes(activeTask))
  : undefined;
```

This caused false positives — for example, if `activeTask` was `kn`,
a `taskRun.metadata.name` containing `tkn` would also match.

---

###  Fix

Replaced the above logic with a label-based match for accurate taskRun resolution:

```ts
const activeTaskRun = taskRuns.find(
  (taskRun) =>
    taskRun.metadata?.labels?.[TektonResourceLabel.pipelineTask] === activeTask,
);
```

This ensures exact matching for both **regular** and **finally** tasks.

---

###  Results

* Log viewer correctly auto-selects the task tab when accessed via `?taskName=<task>`.
* Works for both **PipelineRun** and **TaskRun** views.
* Matches behavior of regular tasks.

---

###  Testing

* Verified with pipelines containing both normal and finally tasks.
* Confirmed that URL navigation now selects the correct tab.
* Unit tests updated to cover finally task selection.

---

###  Additional Info

https://github.com/user-attachments/assets/29d52ee1-d14e-45c3-8513-cc1c7045b382

### DIFF
--- a/src/components/pipelineRuns-details/PipelineRunLogs.tsx
+++ b/src/components/pipelineRuns-details/PipelineRunLogs.tsx
@@ -77,7 +77,11 @@ class PipelineRunLogsWithTranslation extends React.Component<
 
   getActiveTaskRun = (taskRuns: TaskRunKind[], activeTask: string): string => {
     const activeTaskRun = activeTask
-      ? taskRuns.find((taskRun) => taskRun.metadata.name.includes(activeTask))
+      ? taskRuns.find(
+          (taskRun) =>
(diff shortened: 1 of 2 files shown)


logonoff (approved):
/lgtm

PR #771, opened 2025-11-03
SRVKP-8835: updated the sort method for pipeline run and taskrun

anwesha-palit-redhat:
### **Fix TaskRun Sorting by Duration**

**Description of problem:**
When sorting **TaskRuns** by duration, the list was incorrectly sorted based on **completion time** instead of the actual elapsed duration.
This caused incorrect ordering — for example:

* “6 minutes 5 seconds” appearing **above** “50 seconds”
* “3 minutes 10 seconds” appearing **below** “1 minute 5 seconds”

**Root cause:**
The sorting logic used the TaskRun’s `completionTime` value rather than the computed duration field.

**Fix implemented:**

* Updated the TaskRun list to sort by **numeric duration** (elapsed time in seconds) instead of completion time.

**Steps to reproduce:**

1. Have multiple TaskRuns with a wide range of durations (seconds → several minutes).
2. Observe that the TaskRuns are sorted incorrectly by completion time instead of elapsed duration.

**Expected result:**
TaskRuns should be sorted based on the actual numeric duration (in seconds).

**Actual result:**
Durations were sorted lexicographically by string/completion time, resulting in incorrect ordering.

**Please note:**
Please note there is a flag `additional: true` to the TaskRun column

**Screen recording:**
TaskRun duration sorting preview
https://github.com/user-attachments/assets/e4d4eadd-bd73-48a3-9af4-fd7aaf3fa06f

PipelineRun  duration sorting preview
https://github.com/user-attachments/assets/a11f76fc-947d-455b-a7e6-2b6cc0c60d0e

### DIFF
--- a/src/components/pipelineRuns-list/usePipelineRunsColumns.ts
+++ b/src/components/pipelineRuns-list/usePipelineRunsColumns.ts
@@ -8,9 +8,8 @@ import {
   RepositoryLabels,
 } from '../../consts';
 import { PipelineRunKind } from '../../types';
+import { sortPipelineAndTaskRunsByDuration } from '../pipelines-details/pipeline-step-utils';
 import { tableColumnClasses } from './PipelineRunsRow';
-import { sortPipelineRunsByDuration } from '../pipelines-details/pipeline-step-utils';
-
 const usePipelineRunsColumns = (
   namespace: string,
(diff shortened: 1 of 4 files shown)


vikram-raj (commented):
/lgtm

/approve

PR #1024, opened 2026-04-06
SRVKP-10306: Upgrade react-router from v5 to v7 to align with console

anwesha-palit-redhat:
## Summary

- Replaced `react-router 5.3.x`, `react-router-dom 5.3.x`, and `react-router-dom-v5-compat ^6.11.2` with `react-router ~7.13.1` to match the console's version.
- Updated `@openshift-console/dynamic-plugin-sdk{,-internal,-webpack}` from `4.22.0-prerelease.1` to `4.22.0-prerelease.2` (required — the older SDK validates shared modules against v5 and rejects v7).
- Updated `react-i18next` (`~15.1.4` → `~16.5.8`), `react-redux` (`8.1.3` → `~9.2.0`), and `@types/react-redux` (`6.0.2` → `^7.1.34`) to satisfy the new SDK's peer dependencies.
- Migrated all ~67 source files from `react-router-dom-v5-compat` / `react-router-dom` imports to `react-router`.
- Removed the `history` package dependency from `router.ts` utilities, replacing it with native `window.history` APIs.

## Why add `__mocks__/textEncoderDecoder.ts` setup file 

React Router v7 internally uses `TextEncoder` / `TextDecoder` (for cookie and URL handling). The jsdom test environment doesn't provide these globals, so importing `react-router` in any test causes `ReferenceError: TextEncoder is not defined`. The fix adds a `__mocks__/textEncoderDecoder.ts` setup file — matching the exact same approach used by the console project — that polyfills these globals from Node's `util` module before tests run. It also adds `react-router`, `cookie`, and `set-cookie-parser` to `transformIgnorePatterns` exceptions so `babel-jest` can transform their ESM source.

## Why the change in `src/components/utils/router.ts`

The old version of this file imported `createBrowserHistory`, `createMemoryHistory`, and the `History` type from the history npm package (v4). That package was a standalone dependency of `react-router v5` — it provided the browser history abstraction that react-router v5 was built on.
With `react-router v7`, the history package no longer exists as a separate dependency. Its functionality was absorbed into react-router's internals.

**What the old code did:**
Created a History instance via createBrowserHistory() (or createMemoryHistory() in tests), configured with the console's base path.

- Monkey-patched history.push and history.replace to strip the base path from URLs before navigating.
- Used history.replace() inside all the query parameter helpers (setQueryArgument, removeQueryArgument, etc.) to update the URL without a full page reload.

**What the new code does:**
It replaces the history package with the native browser window.history API:

- history.push → window.history.pushState()
- history.replace → window.history.replaceState()
- history.back → window.history.back()

The base-path stripping monkey-patch was also removed. It's no longer needed because callers of history.push (like `handlePipelineRunSubmit` in pipelines-actions.tsx) pass fully-qualified paths from `resourcePathFromModel`(), and the query parameter helpers only modify the query string of the current URL (they read window.location.href directly), so base path handling is irrelevant to them.

## Test plan

- [x] `yarn build` passes
- [x] `yarn test` passes (47/47 suites, 1 skipped — unchanged from before)

## Screen Recording for verification in OCP 4.22 

https://github.com/user-attachments/assets/4e47f90e-92b5-49df-a4a9-9403c1caefde

https://github.com/user-attachments/assets/2d03eab5-a19c-4af6-b96d-b0275b9e2218

https://github.com/user-attachments/assets/8678c3d8-44cc-4980-a02a-9631c735654a

### DIFF
--- a/src/components/multi-tab-list/MultiTabListPage.tsx
+++ b/src/components/multi-tab-list/MultiTabListPage.tsx
@@ -1,13 +1,17 @@
 import type { ReactNode, FC } from 'react';
-import { ActionList, ActionListGroup, ActionListItem, Button } from '@patternfly/react-core';
+import {
+  ActionList,
+  ActionListGroup,
+  ActionListItem,
+  Button,
+} from '@patternfly/react-core';
 import { useTranslation } from 'react-i18next';
-import { Link, useParams } from 'react-router-dom-v5-compat';
+import { Link, useParams, useNavigate } from 'react-router';
 import {
   MenuActions,
   MenuAction,
   SecondaryButtonAction,
 } from './multi-tab-list-page-types';
-import { useNavigate } from 'react-router-dom-v5-compat';
 import { getReferenceForModel } from '../pipelines-overview/utils';
 import {
   HorizontalNav,
--- a/__mocks__/textEncoderDecoder.ts
+++ b/__mocks__/textEncoderDecoder.ts
@@ -1,0 +1,14 @@
+// See https://github.com/jsdom/jsdom/issues/2524
+
+import { TextEncoder, TextDecoder } from 'util';
+
+if (
+  typeof global.TextEncoder !== 'undefined' ||
+  typeof global.TextDecoder !== 'undefined'
+) {
+  throw new Error(
+    'Hello future me. Remove __mocks__/textEncoderDecoder.ts now please',
+  );
+}
+
+Object.assign(global, { TextDecoder, TextEncoder });
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -129,7 +129,11 @@ const config: Config.InitialOptions = {
   // runner: "jest-runner",
 
   // The paths to modules that run some code to configure or set up the testing environment before each test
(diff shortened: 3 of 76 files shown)


src/components/multi-tab-list/MultiTabListPage.tsx
```diff
@@ -1,13 +1,13 @@
...
 import { useTranslation } from 'react-i18next';
-import { Link, useParams } from 'react-router-dom-v5-compat';
+import { Link, useParams } from 'react-router';
 import {
   MenuActions,
   MenuAction,
   SecondaryButtonAction,
 } from './multi-tab-list-page-types';
-import { useNavigate } from 'react-router-dom-v5-compat';
+import { useNavigate } from 'react-router';
```
arvindk-softwaredev: nit: club imports from `react-router` in line #4 and #10

anwesha-palit-redhat: done

arvindk-softwaredev (approved):
/lgtm

PR #1026, opened 2026-04-07
SRVKP-10306: Fixed task node status in pipeline run topology which was not updating in real time

anwesha-palit-redhat:
## Summary

Pipeline topology task nodes were not reflecting live TaskRun status updates — statuses only appeared after hovering on a node or switching tabs. This was caused by a dual MobX instance problem: `PipelineTaskNode` imported `observer` from the plugin-bundled `mobx-react`, while `@patternfly/react-topology` (a shared module loaded from the host console) uses its own MobX instance. The mismatched instances meant observable changes on topology elements were invisible to the plugin's `observer`. 
Additionally, the task status derivation logic in `appendPipelineRunStatus` had issues with mutation, a deprecated Tekton v0 field lookup, and incorrect failure defaults. The `useTRRuns` hook also violated React's rules of hooks with a conditional early return before hook calls.

## Changes

- **`PipelineTaskNode.tsx`**: Import `observer` from `@patternfly/react-topology` instead of `mobx-react` to use the host console's MobX instance, ensuring topology observable changes trigger re-renders. This essentially fixed the main issue we were encountering. Removed redundant `memo` wrapper.

- **`pipeline-utils.ts` (`appendPipelineRunStatus`)**: Replaced `_.merge` with immutable object spreads to prevent stale status persisting via direct mutation of `pipeline.spec.tasks`. 

Removed deprecated `pipelineRun.status.taskRuns` lookup (not present in Tekton v1 — status is now sourced from individual `TaskRun` objects resolved via `childReferences`). 

Fixed fallback logic so tasks default to `Pending` instead of `Failed` when TaskRuns haven't been fetched yet; tasks are only marked `Failed` when the PipelineRun itself has failed.

- **`useTektonResults.ts` (`useTRRuns`)**: Moved the `!isTektonResultEnabled` early return below all hook calls to comply with React's rules of hooks. The guard is now applied inside the `useEffect` condition instead.

## Screen Recording Before

https://github.com/user-attachments/assets/29e3ef89-9582-4d50-9f34-f80367bce2b3

## Screen Recording After

https://github.com/user-attachments/assets/388a4ce5-2de2-411f-acde-a8cb44b69478

### DIFF
--- a/src/components/pipeline-topology/PipelineTaskNode.tsx
+++ b/src/components/pipeline-topology/PipelineTaskNode.tsx
@@ -1,5 +1,5 @@
 import type { FunctionComponent } from 'react';
-import { useRef, useMemo, memo } from 'react';
+import { useRef, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Tooltip } from '@patternfly/react-core';
 import {
@@ -15,9 +15,9 @@ import {
   WhenDecorator,
   WithContextMenuProps,
   WithSelectionProps,
+  observer,
 } from '@patternfly/react-topology';
 import classNames from 'classnames';
-import { observer } from 'mobx-react';
 import { Link } from 'react-router';
 import { NodeType } from './const';
 import { PipelineRunModel, TaskModel } from '../../models';
@@ -223,4 +223,4 @@ const PipelineTaskNode: FunctionComponent<PipelineTaskNodeProps> = ({
   );
 };
 
-export default memo(observer(PipelineTaskNode));
+export default observer(PipelineTaskNode);
--- a/src/components/utils/__tests__/pipeline-utils.spec.ts
+++ b/src/components/utils/__tests__/pipeline-utils.spec.ts
@@ -320,7 +320,7 @@ describe('pipeline-utils ', () => {
     const testTaskRuns = Object.keys(
       pipelineRunWithoutStatus.status.taskRuns,
     ).map((trName) => ({
-      apiVersion: 'v1alpha1',
+      apiVersion: 'v1',
       kind: 'TaskRun',
       metadata: {
         labels: {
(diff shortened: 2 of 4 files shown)


src/components/utils/__tests__/pipeline-utils.spec.ts
```diff
@@ -332,15 +332,12 @@ describe('pipeline-utils ', () => {
       spec: {},
       pipelineTaskName:
         pipelineRunWithoutStatus.status.taskRuns[trName].pipelineTaskName,
-      status: pipelineRunWithoutStatus.status.taskRuns[trName].status,
+      status: _.omit(pipelineRunWithoutStatus.status.taskRuns[trName].status, [
```
anwesha-palit-redhat, line 335: The field `PipelineRun.Status.TaskRuns` is deprecated, I made changes here for the tests to pass

src/components/pipeline-topology/PipelineTaskNode.tsx
```diff
@@ -15,9 +15,9 @@ import {
   WhenDecorator,
   WithContextMenuProps,
   WithSelectionProps,
+  observer,
 } from '@patternfly/react-topology';
 import classNames from 'classnames';
```
anwesha-palit-redhat, line 20: This is essentially the fix for the live status of the task nodes, this module is internally used the PF topology not sure why was it directly used here

src/components/utils/pipeline-utils.ts
```diff
@@ -167,16 +170,10 @@ export const appendPipelineRunStatus = (
     );
     const taskStatus: TaskRunStatus = taskRun?.status;
 
-    const mTask = _.merge(task, {
-      status: pipelineRun?.status?.taskRuns
```
anwesha-palit-redhat, line 171: `pipelineRun.status.taskRuns` is deprecated, therefore removed this

src/components/utils/pipeline-utils.ts
```diff
@@ -150,14 +150,17 @@ export const appendPipelineRunStatus = (
...
-        return _.merge(task, { status: { reason: ComputedStatus.Cancelled } });
+        return { ...task, status: { reason: ComputedStatus.Cancelled } };
       }
       if (
         pipelineRun.spec.status === SucceedConditionReason.PipelineRunPending
       ) {
-        return _.merge(task, { status: { reason: ComputedStatus.Idle } });
+        return { ...task, status: { reason: ComputedStatus.Idle } };
       }
-      return _.merge(task, { status: { reason: ComputedStatus.Failed } });
```
anwesha-palit-redhat, line 160: other than mutating it, this was causing the default behavior of task nodes to be failed, so updated the condition so that it is pending by default, and it shows failed when it actually fails

anwesha-palit-redhat:
closing this PR as the root branch is incorrect

PR #1028, opened 2026-04-09
SRVKP-11533: included missing z-index for TaskSideBar

arvindk-softwaredev:
### **Summary**
Fixes a CSS regression in the Pipeline Builder's TaskSidebar that was introduced as a side effect of the static plugin deprecation. The z-index property for the .opp-task-sidebar element was present in the static plugin's stylesheet but was not carried over to the dynamic plugin's SCSS, causing the sidebar to render behind other page elements.

### **Changes**

- Adds a missing z-index CSS property to the .opp-task-sidebar element in the Pipeline Builder's TaskSidebar component.
- Without this, the sidebar panel could render behind other absolutely/relatively positioned elements on the page, causing it to be visually obscured or unclickable.

Screen Recording - Before

https://github.com/user-attachments/assets/12648ef7-a731-4b57-9ae1-fde79d1d6ca6

Screen Recording - After Fix

https://github.com/user-attachments/assets/87933968-2a78-4090-a843-f250de11673f


anwesha-palit-redhat (approved):
/lgtm

PR #1035, opened 2026-04-10
SRVKP-11394: conditionally navigate to details page on PLR re-run action

arvindk-softwaredev:
### **Summary**

- When a user triggers the "Rerun" action on a PipelineRun, the app now only navigates to the new PipelineRun's details page if the user is already on a PipelineRun details page. Previously, the re-run action always navigated to the new run's details page regardless of the current context (from the list page), which was disruptive.
- Added a loading state to PipelineRunDetails to gracefully handle the brief moment when the pipelineRun object is not yet available (e.g., after navigation to a newly created run).

### **Changes**

1. `pipelineRun-actions-provider.tsx `
     Uses useLocation to detect if the user is on a PipelineRun details page; conditionally navigates to the new run's details only when re-running from the details view
     
2. `PipelineRunDetails.tsx` - Returns a Loading spinner when pipelineRun is not yet available, preventing render errors during navigation

### **Screen Recording Before**

https://github.com/user-attachments/assets/14a2517b-a9f8-414b-95c7-59018c26b05b

### **Screen Recording After**

https://github.com/user-attachments/assets/6cf4750e-b5d7-4a2c-8a9b-f18fce0bf4a1

### DIFF
--- a/src/components/pipelineRuns-details/PipelineRunDetails.tsx
+++ b/src/components/pipelineRuns-details/PipelineRunDetails.tsx
@@ -6,6 +6,7 @@ import { PipelineRunModel } from '../../models';
 import PipelineRunVisualization from './PipelineRunVisualization';
 import PipelineRunCustomDetails from './PipelineRunCustomDetails';
 import { Grid, GridItem, PageSection, Title } from '@patternfly/react-core';
+import { Loading } from '../Loading';
 
 type PipelineRunDetailsProps = {
   obj: PipelineRunKind;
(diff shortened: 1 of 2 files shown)


anwesha-palit-redhat (approved):
/lgtm

src/components/pipelineRuns-details/PipelineRunDetails.tsx
```diff
@@ -15,8 +16,12 @@ const PipelineRunDetails: FC<PipelineRunDetailsProps> = ({
   obj: pipelineRun,
 }) => {
   const { t } = useTranslation('plugin__pipelines-console-plugin');
+  if (!pipelineRun) {
```
anwesha-palit-redhat, line 19: 👍

PR #1055, opened 2026-05-04
SRVKP-10841,SRVKP-11069,SRVKP-11071,SRVKP-11073: ajv and undici both transitive dependency via resolution

anwesha-palit-redhat:
### Fix CVEs in transitive dependencies (`ajv`, `undici`)

Enforces patched versions of vulnerable transitive dependencies using Yarn `resolutions`.

**Fix:**

```json
"resolutions": {
  "ajv@^6.0.0": "^6.14.0",
  "ajv@^8.0.0": "^8.18.0",
  "undici@^6.0.0": "^6.24.0",
  "undici@^7.0.0": "^7.24.0"
}
```

**Notes:**

- Multiple versions of ajv and undici exist because different transitive dependencies require different major versions (e.g. webpack loaders use ajv@6 while newer plugins use ajv@8)
- These major versions are not backward compatible, so they cannot be deduplicated into a single version without breaking dependencies
- This change upgrades each major line to a patched version while preserving compatibility
- We are upgrading openapi-typescript (direct dependency) to remove undici@5

**Testing:**

* Build passes
image

* Unit tests pass
image

* Basic UI flows verified

https://github.com/user-attachments/assets/fd48177c-d811-44b6-be00-f4289c2556cb

* Patched versions confirmed in dependency tree
image


arvindk-softwaredev (approved):
/lgtm

PR #1089, opened 2026-05-29
SRVKP-12183: added missing label suggestions for CRs

anwesha-palit-redhat:
## Summary

Add label filter suggestions to the DataView filter toolbar, matching the behavior of the OpenShift Console’s `ListPageFilter` / `AutocompleteInput` label filter.

- **`useDataViewFilter`**: Collects unique `key=value` labels from the resource data passed into the hook and exposes them via `filterValues.labelSuggestions`, so list pages do not need any changes.
- **`DataViewFilterToolbar`**: Replaces the plain label `SearchInput` with a `LabelFilterInput` that shows suggestions as users type.
- **Suggestion list behavior**:
  - Suggestions appear only after the user enters text (not on focus alone)
  - Case-insensitive substring matching against the full `key=value` string (matches both key and value)
  - Normalizes whitespace around `=` (e.g. `key = value` → `key=value`)
  - Excludes labels already selected as filter chips
  - Supports click-to-select and Enter to add a custom label
  - Uses PatternFly `SelectList` + purple `Label` chips, styled similarly to the console repo
  - Suggestion box is scrollable with `max-height: 200px`

Here are titles you can add under **Screen recordings**:

## Screen recordings

- **User events**

https://github.com/user-attachments/assets/5912e223-fca6-4c15-ba40-8562aebc9325

- **Cluster Trigger Bindings and Pipelines with label filtering** 

https://github.com/user-attachments/assets/640ebaf5-c2a0-47c9-86ea-0722fec9f36a

- **TaskRuns with pagination** 

https://github.com/user-attachments/assets/f8a85647-ef5f-4676-88c6-7fa67e4b051c

- **Pipelines, PipelineRuns, and Tasks** 

https://github.com/user-attachments/assets/25f0dbaa-8ed5-4519-bd64-27a14ffb3226

## Assisted by:
Claude 4.6 Opus High

### DIFF
--- a/src/components/common/DataViewFilterToolbar.scss
+++ b/src/components/common/DataViewFilterToolbar.scss
@@ -1,0 +1,17 @@
+.pipelines-console-plugin-position-relative {
+  position: relative;
+}
+
+.cp-label-suggestion-box {
+  background-color: var(--pf-t--global--background--color--floating--default);
+  box-shadow: var(--pf-t--global--box-shadow--md);
+  display: flex;
+  flex-direction: column;
+  gap: var(--pf-t--global--spacer--gap--group--vertical);
+  padding: var(--pf-t--global--spacer--sm);
+  max-height: 200px;
+  overflow-y: auto;
+  position: absolute;
+  width: 100%;
+  z-index: var(--pf-t--global--z-index--sm);
+}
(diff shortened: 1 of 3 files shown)


arvindk-softwaredev:
verified using the PR image

https://github.com/user-attachments/assets/84e0618f-521e-4a00-89a9-ea9d1b9b39f9

arvindk-softwaredev (approved):
/lgtm

PR #1092, opened 2026-06-01
SRVKP-9448: Resolve overlapping text in Approvals list Current Status column

adityavshinde:
## Description:
This PR fixes a UI rendering bug in the OpenShift Pipelines Approvals tab where the "Current status" elements were superimposing on top of one another, making the text unreadable.

## Root Cause & Technical Fix:
The issue was caused by a position: fixed CSS property applied to the status component. Because fixed positioning removes elements from the standard document flow, multiple status items within the column were collapsing onto the exact same coordinate space on the screen.

By removing position: fixed, the status components now respect the natural flow of the table cell, restoring proper alignment and readability.

### Before
[SRVKP-9448](https://redhat.atlassian.net/browse/SRVKP-9448)-Before

### After
[SRVKP-9448](https://redhat.atlassian.net/browse/SRVKP-9448)-After

## Release Notes
The "Current Status" column now renders correctly instead of an overlap, making it easier to read and verify pipeline tasks.

### DIFF
--- a/src/components/approval-tasks/ApprovalRow.scss
+++ b/src/components/approval-tasks/ApprovalRow.scss
@@ -1,5 +1,4 @@
 .pipelines-approval-status-icon {
-  position: fixed;
   margin-left: -10px;
 
   @-moz-document url-prefix() {


anwesha-palit-redhat:
Verified
Screenshot 2026-06-04 at 13 10 04

anwesha-palit-redhat (approved):
/lgtm

PR #1104, opened 2026-06-04
Update OWNERS to remove specific users

anwesha-palit-redhat:
Removed vikram-raj and lokanandaprabhu from approvers and reviewers.


arvindk-softwaredev (approved):
/lgtm

PR #1123, opened 2026-06-10
SRVKP-11201: Fixes PipelineRuns/TaskRuns lingering in the list after delete when Tekton Results is enabled.

anwesha-palit-redhat:
## Root cause

Two things were contributing:

1. **Stale cache in `useRuns`** — removed runs were being re-inserted via `differenceBy` / `etcdRunsRef`, so deleted resources could stay visible even after the k8s watch dropped them.

2. **Tekton Results annotation timing** — after delete, `resource.loaded.from.tektonResults` appears almost immediately on TR records, but `resource.deleted.in.k8s` is only added client-side in `decodeValueJson` when the TR record has `deletionTimestamp`. That only happens **after** the TR `/get` API runs on re-render. Until then, TR-sourced rows can be misclassified as `cluster-data` (because `isPipelineRunLoadedFromTektonResults` requires **both** annotations) and remain visible under the default cluster-data filter.

## How delete was tracked

The delete callback lives in **console** (via `useDeleteModal`) — we cannot hook into whether the user clicked Delete vs Cancel from the plugin.

Considered copying the delete modal into console-plugin to refetch TR on delete, but that would **not** cover **pruning** (server-side, no user action). So the fix is in the data layer instead.

## Solution

Updated `useRuns` to track `etcdRuns` count. When the count drops (delete or prune), we force a Tekton Results refetch via `trRefetchKey` in `useTRRuns`. In the screen recording you should see a TR `/get` after the delete API calls, which updates annotations and lets the data-source filter behave correctly.

Also removed the stale etcd cache logic and simplified merge/sort handling.

#### Deleting PipelineRun 

https://github.com/user-attachments/assets/8b239b1d-c28a-4bb2-81f1-b38fe426e14c

#### Deleting TaskRun

https://github.com/user-attachments/assets/535adc9b-49ae-47f1-878e-45594371b638

#### Changes in `useRuns` are working when Pruner is enabled

https://github.com/user-attachments/assets/c52ac870-1ba7-4789-8d16-958a9d913f57

### DIFF
--- a/scripts/install-tekton-pruner.sh
+++ b/scripts/install-tekton-pruner.sh
@@ -1,0 +1,143 @@
+#!/usr/bin/env bash
+#
+# Install Tekton Pruner and configure namespace-level pruning.
+#
+# Usage:
+#   ./scripts/install-tekton-pruner.sh
+#   USER_NS=my-team PRUNER_VERSION=v0.4.0 ./scripts/install-tekton-pruner.sh
+#
+# Requires: oc or kubectl, cluster-admin (for SCC on OpenShift)
+#
+set -euo pipefail
+
+PRUNER_VERSION="${PRUNER_VERSION:-v0.4.0}"
+PRUNER_NS="${PRUNER_NS:-tekton-pipelines}"
+USER_NS="${USER_NS:-osp-ui}"
+RELEASE_URL="https://infra.tekton.dev/tekton-releases/pruner/previous/${PRUNER_VERSION}/release.yaml"
+
+# Namespace-level retention defaults (override via env)
+TTL_SECONDS="${TTL_SECONDS:-60}"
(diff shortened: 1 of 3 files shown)


anwesha-palit-redhat:
/retitle [SRVKP-11201](https://redhat.atlassian.net/browse/SRVKP-11201) Fixes PipelineRuns/TaskRuns lingering in the list after delete when Tekton Results is enabled.

arvindk-softwaredev:
Verified the PR image in Openshift cluster:

**Deleting PLR**

https://github.com/user-attachments/assets/fad26e43-4045-4801-a31c-9f28f740dc3a

**Deleting TR**

https://github.com/user-attachments/assets/14d16225-b0da-45ff-b7a1-bbcb98ec7e9f

arvindk-softwaredev (approved):
/lgtm

PR #1133, opened 2026-06-16
SRVKP-12328: Add date range filter for PipelineRuns list page

adityavshinde:
## Summary

Adds a time range filter to the PipelineRuns list page, allowing users to filter pipeline runs
by time period (Last day, Last week, Last month, Last quarter, Last year).

- Integrated as a standard filter category in the existing filter toolbar (alongside Status, Data Source)
- Uses single-select dropdown behavior (picks one time range at a time)
- Server-side filtering via CEL expression for Tekton Results data
- Client-side filtering for Kubernetes/etcd data
- User's selection is persisted across sessions via `useUserPreference` (stored in ConfigMap per user)
- Added `singleSelect` support to `CheckboxFilterConfig` for reuse by other filters

## Changed files

- **`useDateRangeFilter.ts`** (new) - Hook managing time range state, persistence, and CEL generation
- **`useTektonResult.ts`** - Accepts and forwards `filter` parameter to downstream hooks
- **`useTaskRuns.ts`** - Threads `filter` to Tekton Results API, combines with name filter via `AND()`
- **`DataViewFilterToolbar.tsx`** - Added `singleSelect` mode to `CheckboxFilterInput`
- **`PipelineRunsList.tsx`** - Integrates time range into the filter system

## Screen Recording

https://github.com/user-attachments/assets/e0cd797d-3b6c-496c-bfd4-5c8b193a1e17

## Test plan

- [x] Selecting a time range filters PipelineRuns correctly
- [x] Runs without `startTime` (pending) are preserved
- [x] Filter persists across tab switches and page refreshes
- [x] Works with both Tekton Results enabled and disabled
- [x] Clear filters resets time range to default (Last day)
- [x] No flash of default value on navigation
- [x] Unit tests pass (11 tests)
- [x] TypeScript compiles clean

### DIFF
--- a/src/components/common/DataViewFilterToolbar.tsx
+++ b/src/components/common/DataViewFilterToolbar.tsx
@@ -35,6 +35,7 @@ export interface CheckboxFilterConfig {
   placeholder?: string;
   options: FilterOption[];
   defaultValues?: string[];
+  singleSelect?: boolean;
 }
 
 export interface FilterValues {
@@ -77,6 +78,11 @@ const CheckboxFilterInput: FC<{
 
   const onMenuSelect = (_event: unknown, itemId: string | number) => {
     const id = String(itemId);
+    if (config.singleSelect) {
+      onSelect([id]);
+      setIsOpen(false);
+      return;
+    }
     const newSelected = selected.includes(id)
       ? selected.filter((s) => s !== id)
       : [...selected, id];
@@ -121,7 +127,7 @@ const CheckboxFilterInput: FC<{
               ) : undefined
             }
           >
-            {config.placeholder || config.title}
+            {config.placeholder}
           </MenuToggle>
         }
         triggerRef={toggleRef}
@@ -138,15 +144,17 @@ const CheckboxFilterInput: FC<{
                   <MenuItem
                     key={option.value}
                     itemId={option.value}
-                    hasCheckbox
+                    hasCheckbox={!config.singleSelect}
                     isSelected={selected.includes(option.value)}
                   >
                     {option.label}
-                    <Badge isRead className="pf-v6-u-ml-sm">
-                      {option.totalCount !== undefined
-                        ? `${option.count}/${option.totalCount}`
-                        : option.count}
-                    </Badge>
+                    {!config.singleSelect && (
+                      <Badge isRead className="pf-v6-u-ml-sm">
+                        {option.totalCount !== undefined
+                          ? `${option.count}/${option.totalCount}`
+                          : option.count}
+                      </Badge>
+                    )}
                   </MenuItem>
                 ))}
               </MenuList>
@@ -484,7 +492,10 @@ export const DataViewFilterToolbar: FC<DataViewFilterToolbarProps> = ({
                     config={filterConfig}
                     selected={selected}
                     onSelect={(values) =>
-                      onFilterChange(filterConfig.id, values)
+                      onFilterChange(
+                        filterConfig.id,
+                        filterConfig.singleSelect ? values[0] : values,
+                      )
                     }
                   />
                 </ToolbarFilter>
--- a/src/components/hooks/useDateRangeFilter.ts
+++ b/src/components/hooks/useDateRangeFilter.ts
@@ -1,0 +1,77 @@
+import { useUserPreference } from '@openshift-console/dynamic-plugin-sdk';
+import { useMemo } from 'react';
+import { USER_PREFERENCE_PREFIX } from '../../consts';
+
+const s = 1000;
+const m = 60 * s;
+const h = 60 * m;
+const d = 24 * h;
+const w = 7 * d;
+
+const DURATION_MS: Record<string, number> = { s, m, h, d, w };
+
+/** Parse a duration string like "1d", "1w", "4w 2d" into milliseconds. */
+export const parseDurationForDateRangeFiltering = (
+  duration: string,
+): number => {
+  let total = 0;
+  for (const part of duration.trim().split(/\s+/)) {
+    const match = part.match(/^(\d+)([wdhms])$/);
+    if (!match) return 0;
+    total += parseInt(match[1], 10) * DURATION_MS[match[2]];
+  }
+  return total;
+};
+
+/** Midnight (00:00) of the current day in local time. */
+const startOfToday = (): number => {
+  const now = new Date();
+  now.setHours(0, 0, 0, 0);
+  return now.getTime();
+};
+
+export type DateRangeFilterResult = {
+  timespan: number;
+  setTimespanDateFilter: (ms: number) => void;
+  startDate: number | undefined;
+  dateFilterCEL: string;
+  preferenceLoaded: boolean;
+};
+
+export const NO_DATE_RANGE_FILTER = 0;
+
+// When persistPreference is false, no ConfigMap entry is auto-created; sync is enabled when resource type is defined
+export const useDateRangeFilter = (
+  resourceType: string | undefined,
+  persistPreference = false,
+): DateRangeFilterResult => {
+  const [timespan, setTimespanDateFilter, preferenceLoaded] =
+    useUserPreference<number>(
+      `${USER_PREFERENCE_PREFIX}.dateRangeFilter.${resourceType}`,
+      persistPreference ? NO_DATE_RANGE_FILTER : undefined,
+      persistPreference,
+    );
+
+  const selectedTimespan = timespan ?? NO_DATE_RANGE_FILTER;
+
+  // Normalize to midnight so "Last day" means from 00:00 yesterday, not 24h ago
+  const startDate = useMemo(() => {
+    if (!selectedTimespan) return undefined;
+    return startOfToday() - selectedTimespan;
+  }, [selectedTimespan]);
+
+  const dateFilterCEL = useMemo(() => {
+    if (!startDate) return '';
+    return `data.status.startTime > timestamp("${new Date(
+      startDate,
+    ).toISOString()}")`;
+  }, [startDate]);
+
+  return {
+    timespan: selectedTimespan,
+    setTimespanDateFilter,
+    startDate,
+    dateFilterCEL,
+    preferenceLoaded,
+  };
+};
--- a/src/components/hooks/useTaskRuns.ts
+++ b/src/components/hooks/useTaskRuns.ts
@@ -21,7 +21,7 @@ import {
 } from '../../models';
 import { ApprovalTaskKind, PipelineRunKind, TaskRunKind } from '../../types';
 import { useDeepCompareMemoize } from '../utils/common-utils';
-import { EQ } from '../utils/tekton-results';
+import { AND, EQ } from '../utils/tekton-results';
 import { useMultiClusterProxyService } from './useMultiClusterProxyService';
 import { useMultiClusterTaskRuns } from './useMultiClusterTaskRuns';
 import { useTRRuns } from './useTektonResults';
@@ -191,12 +191,14 @@ export const usePipelineRuns = (
     selector?: Selector;
     limit?: number;
     name?: string;
+    filter?: string;
   },
 ): [PipelineRunKind[], boolean, boolean, Error | undefined, boolean, boolean] =>
   useRuns<PipelineRunKind>(PIPELINE_RUN_GVK, namespace, {
     selector: options?.selector,
     limit: options?.limit /* similar to one present in UseTaskRunsOptions */,
     name: options?.name /* similar to one present in UseTaskRunsOptions */,
+    filter: options?.filter,
   });
 
 export const useRuns = <Kind extends K8sResourceKind>(
@@ -207,6 +209,7 @@ export const useRuns = <Kind extends K8sResourceKind>(
     limit?: number;
     name?: string;
     skipFetch?: boolean;
+    filter?: string; // CEL expression sent to Tekton Results to retrieve PRs within the date range
   },
   pipelineRunFinished?: boolean,
   pipelineRunManagedBy?: string,
@@ -315,10 +318,10 @@ export const useRuns = <Kind extends K8sResourceKind>(
 
   const trOptions: typeof optionsMemo = useMemo(() => {
     if (optionsMemo?.name) {
-      const { name, ...rest } = optionsMemo;
+      const { name, filter, ...rest } = optionsMemo;
       return {
         ...rest,
-        filter: EQ('data.metadata.name', name),
+        filter: AND(EQ('data.metadata.name', name), filter),
       };
     }
     return optionsMemo;
--- a/src/components/hooks/useTektonResult.ts
+++ b/src/components/hooks/useTektonResult.ts
@@ -1,5 +1,5 @@
 import { Selector, useFlag } from '@openshift-console/dynamic-plugin-sdk';
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { RepositoryFields, RepositoryLabels } from '../../consts';
 import { FLAGS, PipelineRunKind } from '../../types';
 import { getTaskRunLog } from '../utils/tekton-results';
@@ -9,7 +9,7 @@ export type GetNextPage = () => void | undefined;
 
 export const useGetPipelineRuns = (
   ns: string,
-  options?: { name: string; kind: string },
+  options?: { name: string; kind: string; dateRangeFilter?: string },
 ): [
   PipelineRunKind[],
   boolean,
@@ -31,12 +31,18 @@ export const useGetPipelineRuns = (
     };
   }
 
-  return usePipelineRuns(
-    ns,
-    selector && {
-      selector,
-    },
-  );
+  const pipelineRunOptions = useMemo(() => {
+    const opts: Record<string, unknown> = {};
+    if (selector) {
+      opts.selector = selector;
+    }
+    if (options?.dateRangeFilter) {
+      opts.filter = options.dateRangeFilter;
+    }
+    return opts;
+  }, [selector, options?.dateRangeFilter]);
+
+  return usePipelineRuns(ns, pipelineRunOptions);
 };
 
 export const useTRTaskRunLog = (
--- a/src/components/pipelineRuns-list/PipelineRunsList.tsx
+++ b/src/components/pipelineRuns-list/PipelineRunsList.tsx
@@ -1,5 +1,5 @@
 import type { FC } from 'react';
-import { useEffect, useMemo, useRef } from 'react';
+import { useMemo, useRef, useEffect } from 'react';
 import { useParams, useSearchParams } from 'react-router';
 import { ListPageBody } from '@openshift-console/dynamic-plugin-sdk';
 import usePipelineRunsColumns from './usePipelineRunsColumns';
@@ -11,6 +11,7 @@ import { ConsoleDataView } from '@openshift-console/dynamic-plugin-sdk-internal'
 import { useTranslation } from 'react-i18next';
 import { useDataViewFilter } from '../hooks/useDataViewFilter';
 import { DataViewFilterToolbar } from '../common/DataViewFilterToolbar';
+import { useDateRangeFilter } from '../hooks/useDateRangeFilter';
 
 import './PipelineRunsList.scss';
 
@@ -48,8 +49,14 @@ const PipelineRunsList: FC<PipelineRunsListProps> = ({
     }
   }, []);
 
+  const { dateFilterCEL } = useDateRangeFilter('PipelineRun');
+
   const [pipelineRuns, k8sLoaded, trLoaded, pipelineRunsLoadError] =
-    useGetPipelineRuns(namespace, { name: PLRsForName, kind: PLRsForKind });
+    useGetPipelineRuns(namespace, {
+      name: PLRsForName,
+      kind: PLRsForKind,
+      dateRangeFilter: dateFilterCEL,
+    });
 
   const {
     filterValues,
--- a/locales/en/plugin__pipelines-console-plugin.json
+++ b/locales/en/plugin__pipelines-console-plugin.json
@@ -203,6 +203,7 @@
   "Filter by label": "Filter by label",
   "Filter by name": "Filter by name",
   "Filter by status": "Filter by status",
+  "Filter by time range": "Filter by time range",
   "Finally task": "Finally task",
   "Finally tasks": "Finally tasks",
   "Fit to screen": "Fit to screen",
--- a/src/components/hooks/__tests__/useDateRangeFilter.spec.ts
+++ b/src/components/hooks/__tests__/useDateRangeFilter.spec.ts
@@ -1,0 +1,99 @@
+import { useUserPreference } from '@openshift-console/dynamic-plugin-sdk';
+import { testHook } from '../../../test-data/utils/hooks-utils';
+import {
+  parseDurationForDateRangeFiltering,
+  useDateRangeFilter,
+} from '../useDateRangeFilter';
+
+jest.mock('@openshift-console/dynamic-plugin-sdk', () => ({
+  useUserPreference: jest.fn(),
+}));
+
+const useUserPreferenceMock = useUserPreference as jest.Mock;
+
+const ONE_DAY_MS = 86400000;
+const ONE_WEEK_MS = 604800000;
+
+describe('useDateRangeFilter', () => {
+  let setTimespanDateFilterMock: jest.Mock;
+
+  beforeEach(() => {
+    setTimespanDateFilterMock = jest.fn();
+    useUserPreferenceMock.mockReturnValue([
+      ONE_DAY_MS,
+      setTimespanDateFilterMock,
+    ]);
+  });
+
+  it('should return timespan from user preference', () => {
+    const { result } = testHook(() => useDateRangeFilter('pipelineRun'));
+    expect(result.current.timespan).toBe(ONE_DAY_MS);
+  });
+
+  it('should compute startDate from midnight minus timespan', () => {
+    const { result } = testHook(() => useDateRangeFilter('pipelineRun'));
+    const midnight = new Date();
+    midnight.setHours(0, 0, 0, 0);
+    const expected = midnight.getTime() - ONE_DAY_MS;
+    expect(result.current.startDate).toBe(expected);
+  });
+
+  it('should generate a valid CEL expression', () => {
+    const { result } = testHook(() => useDateRangeFilter('pipelineRun'));
+    expect(result.current.dateFilterCEL).toMatch(
+      /^data\.status\.startTime > timestamp\(".*"\)$/,
+    );
+  });
+
+  it('should default to 0 (no filter) when useUserPreference returns undefined', () => {
+    useUserPreferenceMock.mockReturnValue([
+      undefined,
+      setTimespanDateFilterMock,
+    ]);
+    const { result } = testHook(() => useDateRangeFilter('pipelineRun'));
+    expect(result.current.timespan).toBe(0);
+    expect(result.current.dateFilterCEL).toBe('');
+    expect(result.current.startDate).toBeUndefined();
+  });
+
+  it('should expose setTimespanDateFilter from the preference hook', () => {
+    const { result } = testHook(() => useDateRangeFilter('pipelineRun'));
+    result.current.setTimespanDateFilter(ONE_WEEK_MS);
+    expect(setTimespanDateFilterMock).toHaveBeenCalledWith(ONE_WEEK_MS);
+  });
+
+  it('should use a different preference key for taskRun', () => {
+    const { result } = testHook(() => useDateRangeFilter('taskRun', true));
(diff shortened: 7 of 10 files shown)


src/components/pipelineRuns-list/PipelineRunsList.tsx
```diff
@@ -48,42 +57,113 @@ const PipelineRunsList: FC<PipelineRunsListProps> = ({
...
+    [baseFilterValues, currentKey],
+  );
+
+  const checkboxFilters = useMemo(
+    () => [
+      ...updatedCheckboxFilters,
+      {
+        id: 'timeRange',
+        title: t('Time range'),
+        singleSelect: true,
```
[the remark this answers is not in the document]
adityavshinde: Resolved

anwesha-palit-redhat (commented):
1. Disable tekton-results from tektonconfig and check the behave and attach the screen recording 
2. The screen recording you currently added is not clear - unable to understand the flow of what you are trying to show
3. Add some AI generated unit test cases for this change please

adityavshinde:
> 1. Disable tekton-results from tektonconfig and check the behave and attach the screen recording
> 
>     2. The screen recording you currently added is not clear - unable to understand the flow of what you are trying to show
> 
>     3. Add some AI generated unit test cases for this change please

## 1. Screen recording after disabling tekton-results from tektonconfig

https://github.com/user-attachments/assets/f55f522b-dcb2-43bf-8055-8a0c616f661b

## 2. Steps followed in the screen recording (description)

1. Logged in as `approver-user-1` and set the date range filter to "Last quarter".

2. Switched to `kubeadmin` (cluster admin) to verify the preference was persisted -- admin access is required to list ConfigMaps in the openshift-console-user-settings namespace.

3. Confirmed that the `user-settings-<uid>` ConfigMap for approver-user-1 contains the saved preference key `plugin__pipelines-console-plugin.dateRangeFilter` with the expected value.

## 3. Unit tests
Already added in `useDateRangeFilter.spec.ts` and `PipelineRunsList.spec.tsx`

anwesha-palit-redhat:
> > 1. Disable tekton-results from tektonconfig and check the behave and attach the screen recording
> >    
> >    2. The screen recording you currently added is not clear - unable to understand the flow of what you are trying to show
> >    3. Add some AI generated unit test cases for this change please
> 
> ## 1. Screen recording after disabling tekton-results from tektonconfig
>  RFE_without_tkn-results.mp4 
> ## 2. Steps followed in the screen recording (description)
> 1. Logged in as `approver-user-1` and set the date range filter to "Last quarter".
> 2. Switched to `kubeadmin` (cluster admin) to verify the preference was persisted -- admin access is required to list ConfigMaps in the openshift-console-user-settings namespace.
> 3. Confirmed that the `user-settings-<uid>` ConfigMap for approver-user-1 contains the saved preference key `plugin__pipelines-console-plugin.dateRangeFilter` with the expected value.
> 
> ## 3. Unit tests
> Already added in `useDateRangeFilter.spec.ts` and `PipelineRunsList.spec.tsx`

1. We need to have more data, the purpose of this is to understand if the changes are working for etcd plr
2. Same we need a lot more data to test this - user config change and time range persistance is a part of this change, we also need to ensure that we are showing the right data
3. ACK
4. retitle your PR with the Jira so that the label is added please

src/components/hooks/useTaskRuns.ts
```diff
@@ -326,10 +329,10 @@ export const useRuns = <Kind extends K8sResourceKind>(
 
   const trOptions: typeof optionsMemo = useMemo(() => {
     if (optionsMemo?.name) {
-      const { name, ...rest } = optionsMemo;
+      const { name, filter, ...rest } = optionsMemo;
```
anwesha-palit-redhat, line 321: this needs to be rebased with master branch

src/components/common/DataViewFilterToolbar.tsx
```diff
@@ -32,6 +32,7 @@ export interface CheckboxFilterConfig {
   placeholder?: string;
   options: FilterOption[];
   defaultValues?: string[];
+  singleSelect?: boolean;
```
anwesha-palit-redhat, line 38: what is the purpose of single select

src/components/hooks/useTaskRuns.ts
```diff
@@ -207,6 +209,7 @@ export const useRuns = <Kind extends K8sResourceKind>(
     limit?: number;
     name?: string;
     skipFetch?: boolean;
+    filter?: string;
```
arvindk-softwaredev: > - Server-side filtering via CEL expression for Tekton Results data
> - Client-side filtering for Kubernetes/etcd data

can you capture screen recording with network tab to verify if server side filtering is working for tekton-results api

src/components/common/DataViewFilterToolbar.tsx
adityavshinde: The existing `DataViewFilterToolbar` only supported checkbox-style filters where users can select multiple values (e.g., selecting both "Succeeded" and "Failed" statuses). But for the date range filter, selecting multiple time ranges simultaneously doesn't make sense.

src/components/hooks/useTaskRuns.ts
adityavshinde: Server side filtering is working for tekton-results api but for K8s api we do not have server side filtering.

https://github.com/user-attachments/assets/643a41af-1cd3-4b61-b2b5-7388afc9b34d

src/components/hooks/useTaskRuns.ts
anwesha-palit-redhat: is this rebased yet ?

adityavshinde: Yes, I have rebased this. Now making changes accordingly.

src/components/hooks/useTaskRuns.ts
```diff
@@ -315,10 +318,10 @@ export const useRuns = <Kind extends K8sResourceKind>(
 
   const trOptions: typeof optionsMemo = useMemo(() => {
     if (optionsMemo?.name) {
-      const { name, ...rest } = optionsMemo;
+      const { name, filter, ...rest } = optionsMemo;
       return {
         ...rest,
-        filter: EQ('data.metadata.name', name),
+        filter: AND(EQ('data.metadata.name', name), filter),
```
anwesha-palit-redhat, line 324: what if `filter` is null ?

src/components/hooks/useTektonResult.ts
```diff
@@ -31,12 +31,15 @@ export const useGetPipelineRuns = (
...
-  return usePipelineRuns(
-    ns,
-    selector && {
-      selector,
-    },
+  const pipelineRunOptions = useMemo(
+    () => ({
+      ...(selector && { selector }),
+      ...(options?.filter && { filter: options.filter }),
+    }),
```
anwesha-palit-redhat: can you break this down to if-else so that it is easier to read ?
Also describe how the flow happens from pipelinerun list page to useTaskRuns.ts - try and think about the edge cases, also how it might change in future for taskruns

src/components/hooks/useTaskRuns.ts
anwesha-palit-redhat: also add a single line comment for what the `filter` means here

src/components/hooks/useTaskRuns.ts
anwesha-palit-redhat: what is the expected format for the `filter`

src/components/hooks/useDateRangeFilter.ts
```diff
@@ -0,0 +1,50 @@
...
+  preferenceLoaded: boolean;
+};
+
+const SETTINGS_KEY = 'plugin__pipelines-console-plugin.dateRangeFilter';
+
+export const useDateRangeFilter = (): DateRangeFilterResult => {
+  const isTektonResultEnabled = useFlag(FLAG_PIPELINE_TEKTON_RESULT_INSTALLED);
+  const [timespan, setTimespan, preferenceLoaded] = useUserPreference<number>(
+    SETTINGS_KEY,
+    parsePrometheusDuration('1d'),
```
anwesha-palit-redhat: are we setting 1d as default ?

src/components/pipelineRuns-list/PipelineRunsList.tsx
```diff
@@ -48,42 +57,113 @@ const PipelineRunsList: FC<PipelineRunsListProps> = ({
...
     options: {
       resourceType: 'PipelineRun',
       defaultDataSourceValues: ['cluster-data'],
     },
   });
 
+  const currentKey = formatPrometheusDuration(timespan);
+  const timeRangeOptions = isTektonResultEnabled
+    ? TimeRangeOptions()
+    : TimeRangeOptionsK8s();
```
anwesha-palit-redhat: I do not think we need two separate options here as k8 anyway does not support server side rendering, the one in pipeline overview is added because of prometheus
@arvindk-softwaredev wdyt ?

src/components/pipelineRuns-list/PipelineRunsList.tsx
```diff
@@ -48,42 +57,113 @@ const PipelineRunsList: FC<PipelineRunsListProps> = ({
...
+
+  const filterValues = useMemo(
```
anwesha-palit-redhat: how much are we really saving in terms of performance with this useMemo ?

src/components/pipelineRuns-list/PipelineRunsList.tsx
```diff
@@ -48,42 +57,113 @@ const PipelineRunsList: FC<PipelineRunsListProps> = ({
...
     if (bothOrNone) return k8sLoaded && trLoaded;
     if (selectedSources.includes('cluster-data')) return k8sLoaded;
     return trLoaded;
-  }, [k8sLoaded, trLoaded, filterValues?.dataSource]);
+  }, [preferenceLoaded, k8sLoaded, trLoaded, baseFilterValues?.dataSource]);
 
   return (
     <ListPageBody>
-      {!hideTextFilter && (
+      {!hideTextFilter && preferenceLoaded && (
```
anwesha-palit-redhat: what happens when `preferenceLoaded` is false ? the user sees nothing in this implementation, we should add necessary error banners / loaders

src/components/hooks/useDateRangeFilter.ts
```diff
@@ -0,0 +1,50 @@
...
+  dateFilterCEL: string;
+  isTektonResultEnabled: boolean;
+  preferenceLoaded: boolean;
+};
+
+const SETTINGS_KEY = 'plugin__pipelines-console-plugin.dateRangeFilter';
+
+export const useDateRangeFilter = (): DateRangeFilterResult => {
+  const isTektonResultEnabled = useFlag(FLAG_PIPELINE_TEKTON_RESULT_INSTALLED);
+  const [timespan, setTimespan, preferenceLoaded] = useUserPreference<number>(
```
anwesha-palit-redhat: what if this api fails ?
can we bubble the error please

src/components/hooks/useDateRangeFilter.ts
```diff
@@ -0,0 +1,50 @@
+import { useMemo } from 'react';
+import {
+  useFlag,
+  useUserPreference,
+} from '@openshift-console/dynamic-plugin-sdk';
+import { FLAG_PIPELINE_TEKTON_RESULT_INSTALLED } from '../../consts';
+import { parsePrometheusDuration } from '../pipelines-overview/dateTime';
```
anwesha-palit-redhat: why are we using this `parsePrometheusDuration` ?
we are going to query tektonresults and this is very unreadable 
`try {
    const parts = duration
      .trim()
      .split(/\s+/)
      .map((p) => p.match(/^(\d+)([wdhms])$/));
    return _.sumBy(parts, (p) => parseInt(p[1], 10) * units[p[2]]);
  } catch (ignored) {
    // Invalid duration format
    return 0;
  }`

src/components/hooks/useTaskRuns.ts
adityavshinde: > what if `filter` is null ?

`AND()` is a utility function in `tekton-results.ts` that joins CEL expressions with `&&`. It skips falsy values (null, undefined, empty string). So if filter is null:

1. `AND(EQ('data.metadata.name', name), null)` just returns `EQ('data.metadata.name', name)`
2. It behaves exactly like the original code before the change: `filter: EQ('data.metadata.name', name)`

adityavshinde: > what is the expected format for the `filter`

It's a CEL string for the Tekton Results API. The format looks like:
`data.status.startTime > timestamp("2026-06-22T10:00:00.000Z")`

This is generated by `useDateRangeFilter` in dateFilterCEL. It's the same format used everywhere else in the codebase for Tekton Results queries (e.g., `EQ('data.metadata.name', name)` produces `data.metadata.name == "xyz"`).

src/components/hooks/useDateRangeFilter.ts
adityavshinde: > are we setting 1d as default ?

Yes

src/components/pipelineRuns-list/PipelineRunsList.tsx
adityavshinde: > how much are we really saving in terms of performance with this useMemo ?

Without `useMemo`, React creates a new `filterValues` object on every render, even if nothing changed. `DataViewFilterToolbar` sees a "new" object each time and re-renders unnecessarily.

With `useMemo`, React reuses the same object until `baseFilterValues` or `currentKey` actually changes. This avoids those extra re-renders. Removing it wouldn't cause any visible performance issue as the object is small.

src/components/hooks/useDateRangeFilter.ts
adityavshinde: > why are we using this `parsePrometheusDuration` ? we are going to query tektonresults and this is very unreadable `try { const parts = duration .trim() .split(/\s+/) .map((p) => p.match(/^(\d+)([wdhms])$/)); return _.sumBy(parts, (p) => parseInt(p[1], 10) * units[p[2]]); } catch (ignored) { // Invalid duration format return 0; }`

`parsePrometheusDuration` simply converts a human-readable duration string like '1d' into milliseconds (86400000). It's not related to Prometheus or Tekton Results, it's just a utility for parsing duration strings. We use it because:

- The time range dropdown options are stored as duration keys ('1d', '1w', '1m', etc.)
- We need milliseconds for `Date.now() - timespan` calculation
- The Overview page already uses this same function for the same purpose
- It's already in the codebase, so no new dependency

The name is misleading (it's borrowed from Prometheus-style duration format), but the function itself is just string to milliseconds conversion.

anwesha-palit-redhat: @adityavshinde 
This is a request
We do not want AI responses to review comment, write based on what you understand and can verify else there will be no learning from this 
cc : @arvindk-softwaredev

src/components/hooks/useTaskRuns.ts
anwesha-palit-redhat: please add verification

src/components/pipelineRuns-list/PipelineRunsList.tsx
adityavshinde: I will try this case practically and update

src/components/pipelineRuns-list/PipelineRunsList.tsx
anwesha-palit-redhat: please explain how would this change if the values in dependency arrays were props

src/components/pipelineRuns-list/PipelineRunsList.tsx
anwesha-palit-redhat: bubble the error state and then try please

src/components/hooks/useDateRangeFilter.ts
adityavshinde: > what if this api fails ? can we bubble the error please

If this api fails, then it falls back to the default value which is "Last day". This will not throw an error as it is defined in SDK

src/components/pipelineRuns-list/PipelineRunsList.tsx
adityavshinde: I think it will work same because `useMemo` only compares current values with the previous one. If the parent passes new object every render then it will recompute again and again.

src/components/pipelineRuns-list/PipelineRunsList.tsx
adityavshinde: If we set `preferenceLoaded` as false then it will load forever. Also error is not defined in the SDK, it only returns a tuple of 3 (user preference value, setter callback and loaded boolean). Added a screenshot of this and screen recording for the behavior.


https://github.com/user-attachments/assets/209b6ec0-d0cf-45bf-a057-64c1dded0395

Screenshot From 2026-06-22 22-28-56

src/components/hooks/useTaskRuns.ts
adityavshinde: 1. This verifies that if filter is null, nothing breaks and giving output as before.

https://github.com/user-attachments/assets/cd648381-5a46-4945-8dc1-97929b6b35b2



2. In `summary-api` same pattern has been used.
image

src/components/hooks/useDateRangeFilter.ts
adityavshinde: Should I import the unit constants directly from `dateTime.ts`, then further do the calculation for the startDate?

src/components/hooks/useDateRangeFilter.ts
anwesha-palit-redhat: suppressing the error makes it harder to debug when an issue comes in later, so we will need to change this

src/components/hooks/useDateRangeFilter.ts
anwesha-palit-redhat: you can create your own helper method which is more readable and maintainable

src/components/pipelineRuns-list/PipelineRunsList.tsx
anwesha-palit-redhat: > I think it will work same because useMemo only compares current values with the previous one. If the parent passes new object every render then it will recompute again and again.

exactly, therefore is there a benefit ?

src/components/hooks/useDateRangeFilter.ts
anwesha-palit-redhat: > We need milliseconds for Date.now() - timespan calculation 

We should update this probably, if we do this what essentially happens is it does not compute the whole of the earlier day, example
2 PM on June 23rd - and timespan = 1 day = it will compute till 2 PM of June 22nd, it won't share the whole data for June 22nd, this can misconstrue things. For this filter we should have all data rom now to 00:00 and then add the timespan calculation so that it is normalised. 
wdyt @arvindk-softwaredev ?

src/components/hooks/useDateRangeFilter.ts
anwesha-palit-redhat: `console/frontend/packages/console-shared/src/hooks/useUserPreference.ts`
they are handling the error in someways, let's check this once and see if we can understand if there's an api error

src/components/hooks/useTaskRuns.ts
adityavshinde: Done

src/components/hooks/useTektonResult.ts
adityavshinde: Changed it to "if-else". Also I am creating a doc for explaining the flow, edge cases and future change.

src/components/pipelineRuns-list/PipelineRunsList.tsx
adityavshinde: Got it. I have removed `useMemo`.

src/components/hooks/useDateRangeFilter.ts
adityavshinde: Also adding this in the doc in detail.

src/components/pipelineRuns-list/PipelineRunsList.tsx
anwesha-palit-redhat: but in this case were the dependencies props ?

src/components/common/DataViewFilterToolbar.tsx
```diff
@@ -110,14 +116,16 @@ const CheckboxFilterInput: FC<{
             onClick={handleToggleClick}
             isExpanded={isOpen}
             isFullWidth
-            icon={<FilterIcon />}
+            icon={config.singleSelect ? undefined : <FilterIcon />}
```
arvindk-softwaredev: why do we need to pass `undefined` for the icon?

src/components/hooks/useDateRangeFilter.ts
anwesha-palit-redhat: update `setTimespan` to something more specific

src/components/pipelineRuns-list/PipelineRunsList.tsx
adityavshinde: Removed TimeRangeOptionsK8

src/components/hooks/useDateRangeFilter.ts
adityavshinde: Changed it accordingly

src/components/common/DataViewFilterToolbar.tsx
adityavshinde: Reverted this back to the filter icon which is already in use.

src/components/hooks/useDateRangeFilter.ts
adityavshinde: Updated `setTimespan` to `setTimespanDateRange`

src/components/hooks/useDateRangeFilter.ts
```diff
@@ -0,0 +1,74 @@
+import { useUserPreference } from '@openshift-console/dynamic-plugin-sdk';
+import { useMemo } from 'react';
+import { USER_PREFERENCE_PREFIX } from '../../consts';
+
+const DURATION_MS: Record<string, number> = {
+  s: 1000,
+  m: 60000,
```
anwesha-palit-redhat: nit: wherever you are using this can you please break this down numerically ? it is easier to debug 
example: 
s= 1000,
m= 60*s,
something like that

src/components/hooks/useDateRangeFilter.ts
anwesha-palit-redhat: I still see this as setTimespan 🤔

src/components/pipelineRuns-list/PipelineRunsList.tsx
```diff
@@ -48,33 +54,106 @@ const PipelineRunsList: FC<PipelineRunsListProps> = ({
...
+            {
+              id: 'timeRange',
+              title: t('Time Range'),
+              placeholder: t('Filter by time range'),
+              singleSelect: true,
+              defaultValues: [],
+              options: Object.entries(timeRangeOptions).map(([key, label]) => ({
+                value: key,
+                label,
+                count: 0,
```
anwesha-palit-redhat: what is 0 here ?

adityavshinde: `count` is a mandatory field for filter options, since I am reusing the checkbox filter. But in this case we don't need a count for radio dropdown also `singleSelect: true` hides the count. So 0 is just a placeholder here. (just to satisfy the type)

src/components/hooks/useDateRangeFilter.ts
ankrsinha: Okay, i have made the changes accordingly for better debugging.

src/components/hooks/useDateRangeFilter.ts
ankrsinha: Now, we have completely replaced it with setTimespanDateFilter.

src/components/pipelineRuns-list/PipelineRunsList.tsx
anwesha-palit-redhat: It is not a good practice to use magic numbers
Could we avoid using a magic number here by introducing a named constant for this value ?

adityavshinde: constant name `COUNT_ZERO` would be fine? or shall I add a comment explaining why?

anwesha-palit-redhat: 🤔  it should be something that explains what the value represents

adityavshinde: Done

src/components/pipelineRuns-list/PipelineRunsList.tsx
adityavshinde: No, they are not

src/components/hooks/useDateRangeFilter.ts
```diff
@@ -0,0 +1,75 @@
...
+
+type PageType = 'pipelineRun' | 'taskRun';
+
+export const useDateRangeFilter = (
+  pageType: PageType,
+): DateRangeFilterResult => {
+  const [timespan, setTimespanDateFilter, preferenceLoaded] =
+    useUserPreference<number>(
+      `${USER_PREFERENCE_PREFIX}.dateRangeFilter.${pageType}`,
+      0,
```
anwesha-palit-redhat: @adityavshinde can you update this number `0` with what it actually represents ?

src/components/hooks/useDateRangeFilter.ts
```diff
@@ -0,0 +1,75 @@
...
+      true,
+    );
+
+  const ts = timespan ?? 0;
```
anwesha-palit-redhat: update the name `ts` to something meaningful

src/components/pipelineRuns-list/PipelineRunsList.tsx
```diff
@@ -48,33 +56,106 @@ const PipelineRunsList: FC<PipelineRunsListProps> = ({
...
+
   const [pipelineRuns, k8sLoaded, trLoaded, pipelineRunsLoadError] =
-    useGetPipelineRuns(namespace, { name: PLRsForName, kind: PLRsForKind });
+    useGetPipelineRuns(namespace, {
+      name: PLRsForName,
+      kind: PLRsForKind,
+      filter: dateFilterCEL,
+    });
+
+  const dateFilteredRuns = useMemo(() => {
```
anwesha-palit-redhat: this should be moved to useDataViewFilter and this should check whether there's pre selected `dataSource` as we do not need this for tektonresults data

src/components/pipelineRuns-list/PipelineRunsList.tsx
```diff
@@ -48,33 +56,106 @@ const PipelineRunsList: FC<PipelineRunsListProps> = ({
...
+      if (!st) return true;
+      return new Date(st).getTime() > startDate;
+    });
+  }, [pipelineRuns, timespan, startDate]);
 
   const {
-    filterValues,
-    onFilterChange,
-    onClearAll,
+    filterValues: baseFilterValues,
```
anwesha-palit-redhat: This needs to be moved inside `console-plugin/src/components/hooks/useDataViewFilter.ts`. 
What you are doing here is mutating methods that the hook exposes using `baseFilterValues`, `baseOnFilterChange` and `baseOnClearAll` which is not a practice that should be followed.

src/components/pipelineRuns-list/PipelineRunsList.tsx
adityavshinde: Moved this to useDataViewFilter, also added the check for tekton results.

src/components/hooks/useDateRangeFilter.ts
adityavshinde: Renamed it to selectedTimespan

src/components/hooks/useDateRangeFilter.ts
adityavshinde: Updated it to `NO_DATE_RANGE_FILTER`

src/components/pipelineRuns-list/PipelineRunsList.tsx
adityavshinde: @anwesha-palit-redhat - Done

src/components/common/DataViewFilterToolbar.tsx
```diff
@@ -121,7 +127,7 @@ const CheckboxFilterInput: FC<{
               ) : undefined
             }
           >
-            {config.placeholder || config.title}
+            {config.placeholder}
```
arvindk-softwaredev, line 130: why is `config.title` removed here, is it intentional?

adityavshinde: It was not needed as this fallback would never be reached(config.title). That's why I removed it as it is redundant.

anwesha-palit-redhat (approved):
/lgtm

anwesha-palit-redhat:
Verification
https://drive.google.com/file/d/1gBOga4S6WmjnYIC2EmvvG30846nyK1xK/view

arvindk-softwaredev (approved):
/lgtm

PR #1144, opened 2026-06-26
SRVKP-12527: ability to set default workspace for a pipeline

arvindk-softwaredev:
### **Summary**

- **Add default workspace support for Pipelines in the Pipeline Builder:** Workspaces defined in the Pipeline Builder form now support selecting a default PersistentVolumeClaim (PVC) as the volume type. The chosen PVC and type are persisted in a new `tekton.dev/default-workspace` annotation on the Pipeline resource, so they are restored when the Pipeline is re-opened in the form editor.
- **Revamp PipelineWorkspaces form layout:** Replaced the old `OptionalableWorkspace` component with an inline multi-column layout that shows Name, Type (dropdown, currently PVC-only), and a PVC selector side by side. Added a `PipelineWorkspaces.scss` stylesheet to adjust button positioning in the new layout.
- **Extend PVCDropdown with isFullWidth prop:** The PVCDropdown component now accepts an optional `isFullWidth` prop to control the dropdown width via a PatternFly utility class (pf-v6-u-w-100), replacing the previous hardcoded dropdown--full-width class.

### **Changes**

- Introduced the `DEFAULT_WORKSPACE_ANNOTATION` constant `tekton.dev/default-workspace` in `consts.ts.`
- Extended `TektonWorkspace` type with optional type and claimName fields to carry workspace volume metadata through the form.
- Updated `convertBuilderFormToPipeline` (`utils.ts`) to serialize workspace defaults into the annotation and strip type/claimName from the spec (keeping the Pipeline CRD clean).
- Updated `convertPipelineToBuilderForm` and `getFormData` (`form-switcher-validation.ts`) to parse the annotation back into form values, with graceful fallback on invalid JSON.
- Deleted `OptionalableWorkspace.tsx` (functionality inlined into `PipelineWorkspaces.tsx`).
- Added the "Type" i18n string to the locale file.

### **Test plan**

- [x]  Open the Pipeline Builder, add a workspace, and verify the Name, Type (PVC), and PVC selector columns render correctly
- [x]  Select a PVC for a workspace, save the Pipeline, and confirm the tekton.dev/default-workspace annotation is set on the Pipeline resource
- [x]  Re-open the saved Pipeline in the form editor and confirm the previously selected PVC is pre-populated
- [x]  Toggle the "Optional workspace" checkbox and verify it persists across form/YAML switches
- [x]  Create a Pipeline via YAML without the annotation and switch to form view — workspaces should default to PVC type with no PVC selected

### **ScreenRecording - Part 1**
https://github.com/user-attachments/assets/15059e8e-0f98-459f-8931-b59d6b6ba08f

### **ScreenRecording - Part 2**
https://github.com/user-attachments/assets/6bc3d622-58c7-4e61-b3bc-1b5a742d81df

Assisted-by: Claude 4.6 high


src/components/pipeline-builder/PipelineWorkspaces.tsx
```diff
@@ -17,19 +23,40 @@ const PipelineWorkspaces: FC<PipelineWorkspacesParam> = (props) => {
...
+          isReadOnly={isReadOnly}
+          aria-label={t('Name')}
+        />
+        <DropdownField name="type" items={volumeTypeOptions} disabled />
+        <PVCDropdownForm name="claimName" isFullWidth />
+        <CheckboxField
+          name="optional"
+          label={t('Optional workspace')}
+          className="pf-v6-u-mt-sm"
+        />
```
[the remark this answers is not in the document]
arvindk-softwaredev: addressed in recent commit

src/components/pipeline-builder/utils.ts
```diff
@@ -451,7 +475,9 @@ export const convertBuilderFormToPipeline = (
       ),
       workspaces:
         workspaces.length > 0
-          ? workspaces
+          ? workspaces.map((workspace) =>
+              _.omit(workspace, 'type', 'claimName'),
+            )
           : existingPipeline?.spec?.workspaces ?? [],
```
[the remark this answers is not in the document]
arvindk-softwaredev: addressed in recent commit

src/components/pipeline-builder/utils.ts
```diff
@@ -483,12 +520,23 @@ export const convertPipelineToBuilderForm = (
     },
   } = pipeline;
 
+  let defaultWorkspaceMap: Record<string, { type?: string; name?: string }> =
```
anwesha-palit-redhat: nit: this can be a `const` i think ?

src/components/pipeline-builder/utils.ts
```diff
@@ -483,12 +520,23 @@ export const convertPipelineToBuilderForm = (
...
+    parseDefaultWorkspaceAnnotation(annotations);
+
   return {
     name,
     params,
     workspaces: workspaces.map((workspace) => ({
       ...workspace,
       optional: !!workspace.optional, // Formik fails to understand "undefined boolean" checkbox values
+      type:
```
anwesha-palit-redhat: is this same logic used in a another file ? if it is same then can we use a helper method to abstract the logic in both places ?

src/components/pipeline-builder/utils.ts
```diff
@@ -562,3 +610,22 @@ export const getResourceSidebarSamples = (
 
   return { snippets, samples };
 };
+
+export const parseDefaultWorkspaceAnnotation = (
```
anwesha-palit-redhat, line 609: i think we can check for the annotation here before parsing ?
`if ( annotations &&
    annotations[DEFAULT_WORKSPACE_ANNOTATION]) `

arvindk-softwaredev: added the conditional checks

src/components/pipeline-builder/utils.ts
arvindk-softwaredev: makes sense, abstracted this logic into `extractWorkspaceDefaults` helper method

src/components/pipeline-builder/utils.ts
arvindk-softwaredev: updated to const

src/components/pipeline-builder/utils.ts
```diff
@@ -562,3 +603,38 @@ export const getResourceSidebarSamples = (
...
+  annotations: Record<string, string>,
+): Record<string, { type?: string; name?: string }> => {
+  let defaultWorkspaceMap: Record<string, { type?: string; name?: string }> =
```
anwesha-palit-redhat: `{"sample-workspace":[{"type":"pvc","name":"samplepvc1"}],"sample-workspace2":[{"type":"pvc","name":"samplepvc2"}]}`

if we use this then should we update this type definition ?

src/components/pipeline-builder/utils.ts
```diff
@@ -432,6 +437,30 @@ export const convertBuilderFormToPipeline = (
   const listIds = listTasks.map((listTask) => listTask.name);
   // Strip remaining builder-only properties
   const unhandledSpec = _.omit(others, 'finallyListTasks', 'loadingTasks');
+  let defaultWorkSpaceAnnotation = {};
+  if (workspaces.length > 0) {
+    defaultWorkSpaceAnnotation = workspaces.reduce(
+      (acc: Record<string, Array<object>>, workspace: TektonWorkspace) => {
+        if (workspace.type && workspace.claimName) {
+          let defaultClaim = {
```
anwesha-palit-redhat: nit: can we use const here as well ?

anwesha-palit-redhat:
Verified the changes 

https://github.com/user-attachments/assets/d91dfe51-6569-4a92-805b-bde2e6b58913

anwesha-palit-redhat (approved):
/lgtm

src/components/pipeline-builder/utils.ts
arvindk-softwaredev: updated to const

src/components/pipeline-builder/utils.ts
arvindk-softwaredev: good catch, updated the types in lastes commit

PR #1145, opened 2026-06-29
SRVKP-12532: CVE skill enhancement by preserving yarn.lock during analysis and adding npm ls evidence

ankrsinha:
## Summary
Improve CVE skill accuracy by preserving the committed lockfile during analysis and capturing full dependency evidence.

### Skill (SKILL.md)

- Phase 2a clean install: Changed from **`rm -rf node_modules yarn.lock`** to `rm -rf node_modules dist`  - analysis now runs against the exact dependency versions committed on the branch, preventing false positives from fresh re-resolution
- Resolution strategy: Lockfile regeneration (`rm -rf node_modules yarn.lock && yarn install`) now only happens after explicitly applying resolutions to `package.json`
- Evidence reference: Updated comment templates to reference both `yarnWhyRaw` and `npmLsRaw` fields

### Analysis script (`analyze-deps.ts`)

- Added npmLsRaw field to the script output capturing `npm ls --all <pkg>` output
- Jira comments now include both `yarn why` and `npm` ls evidence instead of only `yarn why`

## Why
Previously the skill removed `yarn.lock` before analysis, which caused Yarn to re-resolve all dependencies from scratch. This could report a CVE as already-remediated when the committed lockfile still pinned the vulnerable version. Additionally, Jira comments only included `yarn why` output, missing the full dependency tree from `npm ls`.

Assisted-by: Claude 4.6 high

### DIFF
--- a/scripts/fix-cves/analyze-deps.ts
+++ b/scripts/fix-cves/analyze-deps.ts
@@ -30,11 +30,15 @@ interface AnalysisResult {
     | 'triage-needed';
   reason: string;
   yarnWhyRaw: string;
+  npmLsRaw: string;
+  /** Per-major-version resolution map, e.g. {"pkg@^2.0.0": "2.5.6", "pkg@^4.0.0": "4.0.6"} */
+  resolutionEntries: Record<string, string>;
 }
 
 interface CLIArgs {
   package: string;
   fixedVersion: string;
+  fixedVersions: string[];
 }
 
 // ---------------------------------------------------------------------------
@@ -65,15 +69,15 @@ function runCmd(cmd: string, args: string[]): string {
 function parseArgs(): CLIArgs {
   const args = process.argv.slice(2);
   let pkg = '';
-  let fixedVersion = '';
+  let fixedVersionRaw = '';
   for (let i = 0; i < args.length; i++) {
     if (args[i] === '--package' && args[i + 1]) pkg = args[++i];
     else if (args[i] === '--fixed-version' && args[i + 1])
-      fixedVersion = args[++i];
+      fixedVersionRaw = args[++i];
   }
-  if (!pkg || !fixedVersion) {
+  if (!pkg || !fixedVersionRaw) {
     console.error(
-      'Usage: analyze-deps.ts --package <name> --fixed-version <ver>',
+      'Usage: analyze-deps.ts --package <name> --fixed-version <ver>[,<ver>,...]',
     );
     process.exit(1);
   }
@@ -81,11 +85,14 @@ function parseArgs(): CLIArgs {
     console.error(`Invalid package name: ${pkg}`);
     process.exit(1);
   }
-  if (!semver.valid(fixedVersion)) {
-    console.error(`Invalid semver version: ${fixedVersion}`);
-    process.exit(1);
+  const fixedVersions = fixedVersionRaw.split(',').map((v) => v.trim());
+  for (const v of fixedVersions) {
+    if (!semver.valid(v)) {
+      console.error(`Invalid semver version: ${v}`);
+      process.exit(1);
+    }
   }
-  return { package: pkg, fixedVersion };
+  return { package: pkg, fixedVersion: fixedVersions[0], fixedVersions };
 }
 
 /**
@@ -148,11 +155,7 @@ function getAllInstalledVersions(pkg: string): string[] {
   return [...versions];
 }
 
-function findVersions(
-  node: any,
-  pkg: string,
-  versions: Set<string>,
-): void {
+function findVersions(node: any, pkg: string, versions: Set<string>): void {
   if (!node || typeof node !== 'object') return;
   if (node.dependencies) {
     for (const [name, dep] of Object.entries<any>(node.dependencies)) {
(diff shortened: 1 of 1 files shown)


anwesha-palit-redhat (commented):
my comment is related to the previous PR https://github.com/openshift-pipelines/console-plugin/pull/1136 so you can add your changes in the current one

we are setting an allowlist there, but never unsetting it, other than manually unsetting the allowlist everytime, is there a command that we can add in the script, which will add the required `npx... etc` when we are using the CVE agent and reset it when we are done ?

ankrsinha:
> my comment is related to the previous PR #1136 so you can add your changes in the current one
> 
> we are setting an allowlist there, but never unsetting it, other than manually unsetting the allowlist everytime, is there a command that we can add in the script, which will add the required `npx... etc` when we are using the CVE agent and reset it when we are done ?

@anwesha-palit-redhat The settings like command allowlist can only be managed through Cursor Settings UI and there is no config file or API that a script can use to modify them. So it is not possible to add this setting and unsetting in the script according to me.

anwesha-palit-redhat:
> > my comment is related to the previous PR #1136 so you can add your changes in the current one
> > we are setting an allowlist there, but never unsetting it, other than manually unsetting the allowlist everytime, is there a command that we can add in the script, which will add the required `npx... etc` when we are using the CVE agent and reset it when we are done ?
> 
> @anwesha-palit-redhat The settings like command allowlist can only be managed through Cursor Settings UI and there is no config file or API that a script can use to modify them. So it is not possible to add this setting and unsetting in the script according to me.

in that case we can prompt the user to approve commands on every use as it would be a much cleaner model

ankrsinha:
> > > my comment is related to the previous PR #1136 so you can add your changes in the current one
> > > we are setting an allowlist there, but never unsetting it, other than manually unsetting the allowlist everytime, is there a command that we can add in the script, which will add the required `npx... etc` when we are using the CVE agent and reset it when we are done ?
> > 
> > 
> > @anwesha-palit-redhat The settings like command allowlist can only be managed through Cursor Settings UI and there is no config file or API that a script can use to modify them. So it is not possible to add this setting and unsetting in the script according to me.
> 
> in that case we can prompt the user to approve commands on every use as it would be a much cleaner model

@anwesha-palit-redhat Yes we can do that, so shall i just update the PR desc of #1136 , since no changes required in this PR for this case ?

anwesha-palit-redhat:
> > > > my comment is related to the previous PR #1136 so you can add your changes in the current one
> > > > we are setting an allowlist there, but never unsetting it, other than manually unsetting the allowlist everytime, is there a command that we can add in the script, which will add the required `npx... etc` when we are using the CVE agent and reset it when we are done ?
> > > 
> > > 
> > > @anwesha-palit-redhat The settings like command allowlist can only be managed through Cursor Settings UI and there is no config file or API that a script can use to modify them. So it is not possible to add this setting and unsetting in the script according to me.
> > 
> > 
> > in that case we can prompt the user to approve commands on every use as it would be a much cleaner model
> 
> @anwesha-palit-redhat Yes we can do that, so shall i just update the PR desc of #1136 , since no changes required in this PR for this case ?

yes please

ankrsinha:
> > > > > my comment is related to the previous PR #1136 so you can add your changes in the current one
> > > > > we are setting an allowlist there, but never unsetting it, other than manually unsetting the allowlist everytime, is there a command that we can add in the script, which will add the required `npx... etc` when we are using the CVE agent and reset it when we are done ?
> > > > 
> > > > 
> > > > @anwesha-palit-redhat The settings like command allowlist can only be managed through Cursor Settings UI and there is no config file or API that a script can use to modify them. So it is not possible to add this setting and unsetting in the script according to me.
> > > 
> > > 
> > > in that case we can prompt the user to approve commands on every use as it would be a much cleaner model
> > 
> > 
> > @anwesha-palit-redhat Yes we can do that, so shall i just update the PR desc of #1136 , since no changes required in this PR for this case ?
> 
> yes please

Done!

scripts/fix-cves/analyze-deps.ts
arvindk-softwaredev, line 409: can we have a condition to add `resolutions` for all the major versions where the CVE is affected

as per [this](https://redhat.atlassian.net/browse/SRVKP-12507) CVE, `form-data` is fixed in 2.5.6, 3.0.5, and 4.0.6 and the release-v1.22.2 branch was having affected versions 2.5.5 and 4.0.4

refer the resolutions added for similar CVE with immutable.js [here](https://github.com/openshift-pipelines/console-plugin/blob/6d5de9be11ae473e07d34e35f38026b5a0a1cbb5/package.json#L192-L193)


**affected versions in release-v1.22.2** 

Image

**commit created by the skill**
Image

ankrsinha: @arvindk-softwaredev @anwesha-palit-redhat I have made the required changes for supporting multiple major fix versions.

anwesha-palit-redhat (approved):
/lgtm

arvindk-softwaredev (approved):
/lgtm

PR #1146, opened 2026-06-29
SRVKP-12032: Issued fix for removing cache and re-triggering tr api on every delete operation

anwesha-palit-redhat:
### AI generated PR description
https://github.com/openshift-pipelines/console-plugin/pull/1146#issuecomment-4831540199

### Screen recordings for PLR 

https://github.com/user-attachments/assets/d4861a4b-d4b3-4d1e-900d-6269d97dbc6a

### Screen recordings for TR 

https://github.com/user-attachments/assets/cc55b70d-7217-486e-8b7a-56d42fc09308

### DIFF
--- a/src/components/hooks/useTaskRuns.ts
+++ b/src/components/hooks/useTaskRuns.ts
@@ -4,9 +4,9 @@ import {
   Selector,
   getGroupVersionKindForModel,
   useFlag,
-  useK8sWatchResource
+  useK8sWatchResource,
 } from '@openshift-console/dynamic-plugin-sdk';
-import { differenceBy, uniqBy } from 'lodash-es';
+import { uniqBy } from 'lodash-es';
 import * as React from 'react';
 import {
   ApprovalFields,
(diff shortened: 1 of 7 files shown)


src/components/pipelines-tasks/TaskRunsRow.tsx
```diff
@@ -66,13 +66,16 @@ const TaskRunKebab: React.FC<TaskRunKebabProps> = ({ obj }) => {
...
-    namespace,
-  });
+  const canDeleteTaskRun =
+    useAccessReview({
+      group: TaskRunModel.apiGroup,
+      resource: TaskRunModel.plural,
+      verb: 'delete',
+      name,
+      namespace,
+    }) && !obj?.metadata?.annotations?.deletionTimestamp
```
arvindk-softwaredev: as `useAccessReview` returns an array and arrays will be truthy,  RBAC value from the hook might be lost
can we refactor this check?

src/components/hooks/useTaskRuns.ts
```diff
@@ -352,8 +342,12 @@ export const useRuns = <Kind extends K8sResourceKind>(
     return value;
   }, [etcdRuns, limit]);
 
-  // cache the last set to identify removed runs
-  etcdRunsRef.current = runs;
+  React.useEffect(() => {
+    if (etcdRuns.length < etcdRunsRef.current.length) {
+      setTrRefreshKey((k) => k + 1 + Date.now());
```
arvindk-softwaredev: nit: why are we adding 1 and timestamp to update `trRefreshKey`

src/components/pipelines-tasks/TaskRunsRow.tsx
anwesha-palit-redhat: nice catch!
updated

src/components/hooks/useTaskRuns.ts
anwesha-palit-redhat: Date.now() should be enough for a refresh key in most cases, just added `+1` to make sure the value changes even if two updates happen in the same millisecond

arvindk-softwaredev:
verified in openshift cluster using the PR image

https://github.com/user-attachments/assets/b71f7e55-3ef2-4e37-903d-036a772dcff5

arvindk-softwaredev (approved):
/lgtm

PR #1147, opened 2026-07-01
SRVKP-12528: Start a PLR with default PVC in workspace

anwesha-palit-redhat:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Migration
- [ ] CVE Fix

## Currently rebased with 
`master`

## Changes in file
`src/components/start-pipeline/utils.ts`

## Summary
If workspace and PVC are selected / set while creating a PL using builder then while starting the PLR those configurations should be selected by default

### Pipeline builder settings

image

### Corresponding values in Start Modal

https://github.com/user-attachments/assets/c9defa6c-9540-48ea-990a-ed8a0e5e2581

### PLR details page ( no code changes made for this )

https://github.com/user-attachments/assets/edad05cc-cf69-44c7-b4e1-19f48bb9a12c

### DIFF
--- a/src/components/start-pipeline/utils.ts
+++ b/src/components/start-pipeline/utils.ts
@@ -40,6 +40,7 @@ import {
   VolumeClaimTemplateType,
 } from '../../types';
 import { ModalErrorContent } from '../modals/error-modal';
+import { parseDefaultWorkspaceAnnotation } from '../pipeline-builder/utils';
 import { getPipelineRunData } from '../utils/utils';
 import { getPipelineRunWorkspaces } from './../utils/pipeline-utils';
 import { CREATE_PIPELINE_RESOURCE } from './validation-utils';
@@ -163,7 +164,12 @@ const supportWorkspaceDefaults =
       type: VolumeTypes.EmptyDirectory,
       data: { emptyDir: {} },
     };
-
+    if (workspace.optional) {
+      workspaceSetting = {
+        type: VolumeTypes.NoWorkspace,
+        data: {},
+      };
+    }
     if (preselectPVC) {
       workspaceSetting = {
         type: VolumeTypes.PVC,
@@ -172,12 +178,6 @@ const supportWorkspaceDefaults =
             claimName: preselectPVC,
           },
         },
-      };
-    }
-    if (workspace.optional) {
-      workspaceSetting = {
-        type: VolumeTypes.NoWorkspace,
-        data: {},
       };
     }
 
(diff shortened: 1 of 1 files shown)


src/components/start-pipeline/utils.ts
anwesha-palit-redhat, line 1: The relevant changes in this PR are in this file

src/components/start-pipeline/utils.ts
```diff
@@ -192,19 +193,43 @@ export const convertPipelineToModalData = (
...
+    if (
+      PVCannotationData[workspaceName] &&
+      PVCannotationData[workspaceName].length > 0
+    ) {
+      const resourceArr = PVCannotationData[workspaceName];
+      const defaultPVC = resourceArr.find(
+        (resource: { type: string; name: string }) =>
+          resource.type === VolumeTypes.PVC,
+      ).name;
+      if (defaultPVC) {
```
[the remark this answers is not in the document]
anwesha-palit-redhat: addressed

src/components/start-pipeline/utils.ts
```diff
@@ -163,7 +164,12 @@ const supportWorkspaceDefaults =
       type: VolumeTypes.EmptyDirectory,
       data: { emptyDir: {} },
     };
-
+    if (workspace.optional) {
```
anwesha-palit-redhat, line 167: Moved this here because the pre-selected PVC was being overwritten when the workspace was optional. This ensures `emptyDir` is set first, and if a default PVC is configured, it overrides `emptyDir` as intended.

src/components/start-pipeline/utils.ts
```diff
@@ -163,7 +164,12 @@ const supportWorkspaceDefaults =
...
+      workspaceSetting = {
+        type: VolumeTypes.NoWorkspace,
+        data: {},
+      };
+    }
     if (preselectPVC) {
       workspaceSetting = {
         type: VolumeTypes.PVC,
```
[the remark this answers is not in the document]
anwesha-palit-redhat: if a `preselectPVC` is present then this preserves this for `optional workspace`, else that would be overwritten

src/components/start-pipeline/utils.ts
arvindk-softwaredev: should we use `if{}..elseif{}` chain instead as we only need to set one `workspaceSetting` at any point?

anwesha-palit-redhat: I don't think we're necessarily setting just one `workspaceSetting` though

The workspace can be optional, have a pre-selected PVC, both, or neither. If we make this an `if...else`, we'd end up having to handle each combination separately (optional + PVC, optional only, PVC only, etc.), whereas keeping them independent seems a bit simpler right ? also didn't add this `if..` i just moved it up in the code block

arvindk-softwaredev: makes sense

arvindk-softwaredev:
verified the changes 

https://github.com/user-attachments/assets/31bfb168-0dad-4dab-8761-7aafd33f90db

arvindk-softwaredev (approved):
/lgtm

PR #1153, opened 2026-07-03
SRVKP-12530: toggle to hide optional params in TaskSidebar

arvindk-softwaredev:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Migration
- [ ] CVE Fix

## Summary

- Adds a "Show required params only" toggle to show/hide optional parameters in the Pipeline Builder's task sidebar allowing users to filter out optional task parameters when configuring tasks. This reduces visual clutter for tasks with many parameters where only a subset are required. 
- When the toggle is on, all the optional parameters will be hidden in the TaskSidebar and only required parameters will be displayed
- If there are no required parameters for the task, an alert message will be displayed.  
- Also, fixed a modal-close bug in the QuickSearch version dropdown by properly using e.stopPropagation().

### Changes

- **Toggle in Pipeline Builder toolbar:** Added a `Switch` control next to the Form/YAML radio group that toggles between showing all parameters and showing only required parameters.
- **Task Sidebar filtering:** When the toggle is enabled, optional parameters (those with a `default` value) are hidden from the task sidebar. If a task has no required parameters, an inline alert is shown instead.
- **QuickSearch modal close fix:** Moved the workaround for the version dropdown closing the QuickSearch modal from a DOM query check in `QuickSearchModalBody` to a proper `e.stopPropagation()` in `PipelineQuickSearchVersionDropdown.onSelect`. This is the correct event-handling pattern and eliminates a fragile DOM traversal.
- **CSS fix:** Added `flex-direction: row` to the inline radio group to prevent layout wrapping when the new toggle is placed alongside it.

### Files changed

| File | Change |
|------|--------|
| `PipelineBuilderForm.tsx` | State for `hideOptionalParam`, passed to sidebar and editor toggle |
| `SyncedEditorField.tsx` | Renders the `Switch` toggle alongside the Form/YAML radio group |
| `TaskSidebar.tsx` | Filters params by required status, shows fallback alert |
| `RadioGroupField.scss` | Layout fix for inline radio + toggle alignment |
| `QuickSearchModalBody.tsx` | Removed DOM-based modal close workaround |
| `PipelineQuickSearchVersionDropdown.tsx` | Added `e.stopPropagation()` in `onSelect` |
| `locales/en/...json` | New i18n strings |

## Test plan

- [ ] Open Pipeline Builder → add a task with both required and optional params
- [ ] Verify the "Show required params only" toggle appears next to the Form/YAML radio
- [ ] Toggle ON → only required params visible in the task sidebar
- [ ] Toggle ON for a task with no required params → info alert shown
- [ ] Toggle OFF → all params visible again
- [ ] Switch to YAML view and back → toggle state persists
- [ ] Open QuickSearch → change task version from dropdown → modal should NOT close

## Screen Recording

https://github.com/user-attachments/assets/2e3d9381-d20b-4ad6-a644-593709a60923


src/components/pipeline-builder/RadioGroupField.scss
```diff
@@ -18,6 +18,7 @@
   &--inline {
     display: flex;
     align-items: center;
+    flex-direction: row !important;
```
arvindk-softwaredev, line 21: note- this is added to align the radio options (Pipeline/Yaml) in a row

locales/en/plugin__pipelines-console-plugin.json
```diff
@@ -323,6 +323,7 @@
   "No parameters are associated with this PipelineRun.": "No parameters are associated with this PipelineRun.",
   "No Properties found": "No Properties found",
   "No requester": "No requester",
+  "No Required Params for this task": "No Required Params for this task",
```
anwesha-palit-redhat: nit: Can we update this to `There are no required params for this task`

src/components/pipeline-builder/SyncedEditorField.tsx
```diff
@@ -200,7 +211,20 @@ const SyncedEditorField: FC<SyncedEditorFieldProps> = ({
...
           isInline
         />
-      </div>
+        <Divider
+          orientation={{
+            default: 'vertical',
+          }}
+        />
+        <div onClick={(e) => e.stopPropagation()}>
+          <Switch
```
anwesha-palit-redhat, line 235: Should we add a tooltip here with a better description which explains that this is for hiding the task params and not pipeline params ?

src/components/pipeline-builder/RadioGroupField.scss
anwesha-palit-redhat: i checked and this class is also present in the console codebase, and they are also doing the same row !important, but to avoid any future conflicts I think we should update the name of the main class `ocs-radio-group-field` wdyt ?

src/components/pipeline-builder/task-sidebar/TaskSidebar.tsx
```diff
@@ -50,6 +52,7 @@ const TaskSidebar: FC<TaskSidebarProps> = (props) => {
     workspaceList,
     selectedData,
     onClose,
+    hideOptionalParam,
```
anwesha-palit-redhat: nit: should we update this name to hideTaskOptionalParam or something that explains that this only applies to tasks and not pipelines ?

arvindk-softwaredev: updated to `hideOptionalTaskParam` in all references

src/components/pipeline-builder/SyncedEditorField.tsx
arvindk-softwaredev: updated the label text to `Show required task params only`

locales/en/plugin__pipelines-console-plugin.json
arvindk-softwaredev: updated the string

src/components/pipeline-builder/RadioGroupField.scss
arvindk-softwaredev: yes, the class from console was not getting applied, probably it is removed in latest versions
updated the classname to `osp-radio-group-field`

arvindk-softwaredev:
@anwesha-palit-redhat 
conditionally disabled the `Show required Task params` only toggle if there are no tasks added from the Pipelinebuilder form or yaml view

https://github.com/user-attachments/assets/963f3d56-1805-49a7-9b28-4204b2a63a83

anwesha-palit-redhat:
Verified using PR image

https://github.com/user-attachments/assets/c764aeda-0b1d-4fef-b682-f5107e492e8a

anwesha-palit-redhat (approved):
/lgtm

PR #149, opened 2024-08-07
SRVKP-5936: Close modal button is not working in dynamic plugin list pages

lokanandaprabhu:
**Issue:**
https://issues.redhat.com/browse/SRVKP-5936

**Analysis / Root cause**: 
onClose function was not submitted to Modal component.

**Solution Description**: 
Added onClose function to Modal component. 

**Screen shots / Gifs for design review**: 
----BEFORE---

https://github.com/user-attachments/assets/211ff1b6-d73a-47b9-b649-f5b9beb19e26

---AFTER---

https://github.com/user-attachments/assets/401d9b34-3637-4b7e-864b-1308dd1074b2

**Unit test coverage report**: 
NA

**Test setup:**

1. Install Pipelines operator and enable dynamic plugin
2. Create a pipeline and click on start in kebab menu
3. Click on X button to close the start pipeline modal

    

**Browser conformance**: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

PR #1154, opened 2026-07-06
SRVKP-12531: Optional param changes for YAML editor

anwesha-palit-redhat:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Migration
- [ ] CVE Fix

## Summary

https://github.com/openshift-pipelines/console-plugin/pull/1154#issuecomment-4914469727

### 10 git-clones in a series

https://github.com/user-attachments/assets/b0a1f016-b0bc-49b9-b071-aba588eb0b16

### light / dark themes for enabled and disabled switch - and switch resets on removing tasks

https://github.com/user-attachments/assets/847a11ce-939c-4f8c-b6ff-b93535adc836

### updating params and switching between YAML and Builder UI

https://github.com/user-attachments/assets/78b257b4-d32b-4b1b-b3ce-c798d6105f72

### updating params and switching between YAML and Builder UI in finally tasks

https://github.com/user-attachments/assets/79dd2c26-fd49-445c-be20-6001e0f50b17

### During review we found an issue due to which we are making the YAML editor read only when the toggle is ON

Root cause: [this](https://github.com/openshift-pipelines/console-plugin/pull/1154/changes/BASE..5ab5353c4e419e9457837964ad34db8ce4033189#diff-9f92a0a0c6fe90f7d44326518e1c521218857b60518f7182093807cdd223821eL95) is causing the issue, the purpose of having `displayValue` was to not mutate the actual `field.value` but `handleOnChange` is mutating it anyway ( it also fires `onSave` and it does not check a specific field that is updated, just puts the newYAML directly ) so the issue  also happens if one edits the YAML and submits ( or switches back to the builder and submits )
We have therefore made the YAML readonly when the toggle mode is turned on ( to prevent this mutation )

https://github.com/user-attachments/assets/2ab9f482-94f7-4f5b-b964-7402d85661c5

### DIFF
--- a/src/components/pipeline-builder/CodeEditorField.tsx
+++ b/src/components/pipeline-builder/CodeEditorField.tsx
@@ -1,7 +1,7 @@
 import type { FC } from 'react';
 import { useRef, useState, useCallback } from 'react';
 import classNames from 'classnames';
-import { Button } from '@patternfly/react-core';
+import { Alert, Button } from '@patternfly/react-core';
 import { InfoCircleIcon } from '@patternfly/react-icons/dist/esm/icons/info-circle-icon';
 import { FormikValues, useField, useFormikContext } from 'formik';
 import { isEmpty } from 'lodash';
@@ -34,6 +34,8 @@ export interface CodeEditorFieldProps extends FieldProps {
   showSamples: boolean;
   showShortcuts?: boolean;
   showMiniMap?: boolean;
+  hideOptionalTaskParam?: boolean;
+  displayValue?: string;
   onSave?: () => void;
 }
 
@@ -53,11 +55,15 @@ const CodeEditorField: FC<CodeEditorFieldProps> = ({
   minHeight,
   onSave,
   language,
+  hideOptionalTaskParam,
+  displayValue,
 }) => {
   const [field] = useField(name);
   const { setFieldValue, setStatus } = useFormikContext<FormikValues>();
   const { t } = useTranslation('plugin__pipelines-console-plugin');
   const editorRef = useRef();
+  const lastDisplayValueRef = useRef<string>();
+  lastDisplayValueRef.current = displayValue;
 
   const [sidebarOpen, setSidebarOpen] = useState<boolean>(true);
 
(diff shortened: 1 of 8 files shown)


src/components/pipeline-builder/SyncedEditorField.tsx
```diff
@@ -144,21 +143,27 @@ const SyncedEditorField: FC<SyncedEditorFieldProps> = ({
...
     setStatus({ submitError: '' });
   };
-
+  /* Reset the toggle when no tasks are present and disable it */
   const isOptionalTaskParamToggleDisabled = useMemo(() => {
     if (editorType === EditorType.Form) {
-      return !formData?.tasks?.length;
+      if (formData?.tasks?.length === 0) {
+        setHideOptionalTaskParam(false);
+        return !formData?.tasks?.length;
```
arvindk-softwaredev: nit: as we have the check in line #149, should we just return `true` here instead of a redundant check

anwesha-palit-redhat: yes yes, i just left it like it was before, will update

src/components/pipeline-builder/CodeEditorField.tsx
```diff
@@ -97,6 +101,8 @@ const CodeEditorField: FC<CodeEditorFieldProps> = ({
     setStatus({ submitError: '' });
   };
 
+  const editorValue = hideOptionalTaskParam ? displayValue : field.value;
```
arvindk-softwaredev: i think this is causing the yaml to save with required params only when the toggle is on, thus leaving behind the optional task params in the actual yaml when creating a pipeline

https://github.com/user-attachments/assets/d506d244-99bd-4fc5-a2ec-f79eb519f2b0

arvindk-softwaredev:
verified the changes with PR image in Openshift cluster

https://github.com/user-attachments/assets/b60ce8a5-de2b-47cd-945a-b393281dcf38

arvindk-softwaredev (approved):
/lgtm

PR #1169, opened 2026-07-10
fix(pipeline-augment): scope PipelineRun match to Pipeline namespace

### LINKED ISSUE #1168: [BUG] Pipelines list "Last run" shows a PipelineRun from a different namespace when multiple namespaces have a Pipeline with the same name
## Summary
When the OpenShift Pipelines console plugin's Pipelines list page is viewed with "All Projects" selected, and multiple namespaces contain a `Pipeline` resource with the same `name`, the "Last run" column shows the same (most recent) `PipelineRun` for all of them — even though that `PipelineRun` only actually exists in one namespace.

## Root cause
`augmentRunsToData()` in `src/components/utils/pipeline-augment.ts` matches `PipelineRun`s to `Pipeline`s using only the `tekton.dev/pipeline` label, without checking that the `PipelineRun`'s namespace matches the `Pipeline`'s namespace:

```ts
export const augmentRunsToData = (
  pipelines: PropPipelineData[],
  pipelineruns: PipelineRunKind[],
): PropPipelineData[] => {
  return pipelines.map((pipeline) => {
    const prsForPipeline = pipelineruns.filter(
      (pr) =>
        pr.metadata.labels?.['tekton.dev/pipeline'] === pipeline.metadata.name,
    );
    pipeline.latestRun = getLatestRun(prsForPipeline, 'creationTimestamp');
    return pipeline;
  });
};
```

In the "All Projects" view, `pipelines` and `pipelineruns` contain resources from every namespace. Since the filter only checks the `tekton.dev/pipeline` label (the Pipeline name), a `Pipeline` named `my-pipeline` in namespace `ns-b` ends up picking up the latest `PipelineRun` named/labeled `my-pipeline` from *any* namespace (e.g. `ns-a`), not just its own.

## Steps to reproduce
1. Create a `Pipeline` with the same name (e.g. `my-pipeline`) in several namespaces (e.g. `ns-a`, `ns-b`, `ns-c`, ...).
2. Trigger a `PipelineRun` for that `Pipeline` in only one of those namespaces (e.g. `ns-a`).
3. Open the OpenShift web console → Pipelines → select "All Projects".
4. Every `my-pipeline` row across all namespaces shows the same "Last run" name/status/time as the one namespace where it actually ran, even though no `PipelineRun` object exists in the other namespaces (verified via `oc get pipelinerun -n <ns>`).

## Expected behavior
Each `Pipeline` row should only reflect `PipelineRun`s that exist in its own namespace.

## Environment
- Red Hat OpenShift on IBM Cloud 4.16.63
- OpenShift Pipelines Operator 1.22.4 (`console-plugin` `release-v1.22.x`)

seiya-koji:
## Summary
`augmentRunsToData()` matched `PipelineRun`s to `Pipeline`s using only the `tekton.dev/pipeline` label, without checking that the `PipelineRun`'s namespace matches the `Pipeline`'s namespace. In the Pipelines list "All Projects" view, this caused every `Pipeline` with the same name across different namespaces to show the same "Last run", even when the `PipelineRun` only existed in one of those namespaces.

Fixes #1168

## Fix
Add a namespace equality check to the existing label match in `augmentRunsToData`.

## Test plan
- Added a regression test in `pipeline-augment.spec.ts` reproducing the bug: two `Pipeline`s with the same name in different namespaces, only one of which has a matching `PipelineRun`. Confirmed it fails before the fix and passes after.
- Adjusted pre-existing test fixtures (`pipeline-augment-test-data.ts`) that had `Pipeline`/`PipelineRun` namespace mismatches, so they keep passing under the new namespace-aware matching.
- Ran the full test suite (49 test files, 594 tests total): 593 passed, 1 skipped (pre-existing, unrelated to this change).
- `eslint` on changed files: no issues.


seiya-koji:
Sorry, I mistakenly set the base to `main` instead of `master`.

PR #1170, opened 2026-07-10
fix(pipeline-augment): scope PipelineRun match to Pipeline namespace

### LINKED ISSUE #1168: [BUG] Pipelines list "Last run" shows a PipelineRun from a different namespace when multiple namespaces have a Pipeline with the same name

seiya-koji:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Migration
- [ ] CVE Fix

## Summary

`augmentRunsToData()` matched `PipelineRun`s to `Pipeline`s using only the `tekton.dev/pipeline` label, without checking that the `PipelineRun`'s namespace matches the `Pipeline`'s namespace.
In the Pipelines list "All Projects" view, this caused every `Pipeline` with the same name across different namespaces to show the same "Last run", even when the `PipelineRun` only existed in one of those namespaces.

Fixes #1168

Fix: added a namespace equality check to the existing label match in `augmentRunsToData`.

Test plan:
- Added a regression test in `pipeline-augment.spec.ts` reproducing the bug: two `Pipeline`s with the same name in different namespaces, only one of which has a matching `PipelineRun`. Confirmed it fails before the fix and passes after.
- Adjusted pre-existing test fixtures (`pipeline-augment-test-data.ts`) that had `Pipeline`/`PipelineRun` namespace mismatches, so they keep passing under the new namespace-aware matching.
- Ran the full test suite: all passing (1 pre-existing skip, unrelated to this change).
- `eslint` on changed files: no issues.

### Before

Screen recordings for before/after.

**Before**: When a `create-table` Pipeline was run in only one namespace, every namespace that had a Pipeline with the same name showed that run — the `create-table-ixlmy2` PipelineRun appeared across all of them, all with the same timestamp (`Jul 10, 2026, 3:23 PM`).

image

**After** this fix, only the namespace where the Pipeline actually ran shows a result — each namespace now shows its own PipelineRun.

image

### DIFF
--- a/src/components/utils/__tests__/pipeline-augment-test-data.ts
+++ b/src/components/utils/__tests__/pipeline-augment-test-data.ts
@@ -24,6 +24,7 @@ export const testData: PipelineAugmentData[] = [
         kind: 'PipelineRun',
         metadata: {
           name: 'apple-1-run1',
+          namespace: 'myproject',
           creationTimestamp: '21-05-2019',
           labels: { 'tekton.dev/pipeline': 'apple1' },
         },
@@ -40,13 +41,13 @@ export const testData: PipelineAugmentData[] = [
       {
         metadata: {
           name: 'apple1',
-          namespace: 'myproject',
+          namespace: 'tekton-pipelines',
         },
       },
       {
         metadata: {
           name: 'apple2',
-          namespace: 'myproject',
+          namespace: 'tekton-pipelines',
         },
       },
     ],
@@ -84,8 +85,42 @@ export const testData: PipelineAugmentData[] = [
         kind: 'PipelineRun',
         metadata: {
           name: 'apple-2-run1',
+          namespace: 'tekton-pipelines',
           creationTimestamp: '31-04-2019',
           labels: { 'tekton.dev/pipeline': 'apple2' },
+        },
+        spec: {},
+        status: {
+          pipelineSpec: { tasks: [] },
+          conditions: [{ type: 'Succeeded', status: 'True' }],
+        },
+      },
+    ],
+  },
+  {
+    pipelines: [
+      {
+        metadata: {
+          name: 'simple-pipeline',
+          namespace: 'ns-a',
+        },
+      },
+      {
+        metadata: {
+          name: 'simple-pipeline',
+          namespace: 'ns-b',
+        },
+      },
+    ],
+    pipelineruns: [
+      {
+        apiVersion: 'tekton.dev/v1',
+        kind: 'PipelineRun',
+        metadata: {
+          name: 'simple-pipeline-run-1',
+          namespace: 'ns-a',
+          creationTimestamp: '2026-07-10T01:23:45Z',
+          labels: { 'tekton.dev/pipeline': 'simple-pipeline' },
         },
         spec: {},
         status: {
--- a/src/components/utils/__tests__/pipeline-augment.spec.ts
+++ b/src/components/utils/__tests__/pipeline-augment.spec.ts
@@ -54,6 +54,25 @@ describe('PipelineAugment test correct data is augmented', () => {
       testData[3].pipelineruns[0].metadata.name,
     );
   });
+
+  it('expect a PipelineRun to only be matched to the Pipeline in the same namespace', () => {
+    const newData = augmentRunsToData(
+      testData[4].pipelines,
+      testData[4].pipelineruns,
+    );
+    const pipelineWithRun = newData.find(
+      (p) => p.metadata.namespace === 'ns-a',
+    );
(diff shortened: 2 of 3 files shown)


anwesha-palit-redhat:
Hi @seiya-koji 👋 
Can you please add before / after screen recordings if possible ?

seiya-koji:
@anwesha-palit-redhat 
Added before/after screenshots to the PR description.

seiya-koji:
FYI — this looks like an [openshift/release](https://github.com/openshift/release)-side issue, but flagging it here.

While testing this locally, `yarn run start-console` failed with `useSearchParams`/`useNavigate is not a function`. 
It looks like [`quay.io/openshift/origin-console:latest`](https://github.com/openshift-pipelines/console-plugin/blob/35493d8872ab36b2e4721ccf6acc8237f2f641c2/start-console.sh#L5) is stuck on a 2026-02-14 build (pre-dates react-router v7).

The mirroring job that's supposed to update it has been failing on every run: https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/periodic-image-mirroring-openshift

anwesha-palit-redhat:
Defect verification 

https://github.com/user-attachments/assets/7c8929d2-5409-449b-9baa-afda350e6d5e

Fix verification

https://github.com/user-attachments/assets/7916c9c4-9f82-46b3-adc3-26ed24b11c61

anwesha-palit-redhat (approved):
/lgtm

anwesha-palit-redhat:
@seiya-koji would you please mind cherry-picking this to main ?

seiya-koji:
@anwesha-palit-redhat I'm not too familiar with how this works — does this mean the bot will handle the cherry-pick automatically once this merges, and there's nothing I need to do on my end?
Or would it be better if I opened the PR against main myself?

anwesha-palit-redhat:
> @anwesha-palit-redhat I'm not too familiar with how this works — does this mean the bot will handle the cherry-pick automatically once this merges, and there's nothing I need to do on my end? Or would it be better if I opened the PR against main myself?

my bad, i did not explain it well, we will let this PR merge first
and then we will try the bot cherry-pick, but if it fails then would you be able to manually cherry-pick ? ( this was what i wanted to ask )

seiya-koji:
Got it — if the bot cherry-pick fails, I'll open the PR against main manually.

arvindk-softwaredev (approved):
/lgtm

PR #1172, opened 2026-07-13
SRVKP-12725: Adding util files and updating types for pipeline in pipelines

anwesha-palit-redhat:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Migration
- [ ] CVE Fix

## Summary
These are utility files required for Pipeline-in-Pipeline (PiNP). They're being added separately so they can be reused independently across other components.


src/components/hooks/useChildPipelineRuns.ts
arvindk-softwaredev, line 1: nit: this customhook should be a `.ts` file

anwesha-palit-redhat: done

arvindk-softwaredev (approved):
/lgtm

PR #1191, opened 2026-07-23
SRVKP-13054: Fix close option for the date range filter

adityavshinde:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Migration
- [ ] CVE Fix

## Summary
The date range filter chip's close (X) button was not clearing the filter. When clicking X, deleteChip removes the value from the array producing [], then calls `onFilterChange('timeRange', [])`. Since empty arrays are truthy in JavaScript, the old code `value ? parseDurationForDateRangeFiltering(value as string) : NO_DATE_RANGE_FILTER` would enter the truthy branch and try to parse [] as a string instead of clearing the filter.

### DIFF
--- a/src/components/hooks/useDataViewFilter.ts
+++ b/src/components/hooks/useDataViewFilter.ts
@@ -304,9 +304,10 @@ export const useDataViewFilter = <T extends K8sResourceCommon>({
   const onFilterChange = useCallback(
     (key: string, value: string | string[]) => {
       if (key === 'timeRange') {
+        const selected = Array.isArray(value) ? value[0] : value;
         setTimespanDateFilter(
-          value
-            ? parseDurationForDateRangeFiltering(value as string)
+          selected
+            ? parseDurationForDateRangeFiltering(selected)
             : NO_DATE_RANGE_FILTER,
         );
         resetPage();


src/components/hooks/useDataViewFilter.ts
```diff
@@ -303,10 +303,11 @@ export const useDataViewFilter = <T extends K8sResourceCommon>({
 
   const onFilterChange = useCallback(
     (key: string, value: string | string[]) => {
+      const selected = Array.isArray(value) ? value[0] : value;
```
ankrsinha: Can move the `selected` variable inside the if block of `timeRange`, since it is used inside that block only, just for better readability.

adityavshinde: Resolved

ankrsinha:
The fix looks good to me. Attaching the verification recording

/lgtm

https://github.com/user-attachments/assets/f83ec34d-ecbe-4444-b6fa-7ae0948312fe

src/components/hooks/useDataViewFilter.ts
```diff
@@ -304,9 +304,10 @@ export const useDataViewFilter = <T extends K8sResourceCommon>({
   const onFilterChange = useCallback(
     (key: string, value: string | string[]) => {
       if (key === 'timeRange') {
+        const selected = Array.isArray(value) ? value[0] : value;
```
anwesha-palit-redhat, line 307: nit: Are we expecting value to ever be a standalone string here, or is it always an array (or empty/null)? 

If it's always an array when present, returning `null` as the fallback might be cleaner and avoid returning an unexpected non-array value ? wdyt ?

`const selected = Array.isArray(value) ? value[0] : null;`

anwesha-palit-redhat:
verified and changes are working, left a `nit`

### PipelineRun

https://github.com/user-attachments/assets/d2fb9f44-2e8e-4d49-ab6f-28998e723436

### TaskRun

https://github.com/user-attachments/assets/4269f697-b512-4285-ab8a-e977339949aa

https://github.com/user-attachments/assets/b94b19b6-1d99-4076-abe3-2916adf4fc4f

src/components/hooks/useDataViewFilter.ts
adityavshinde: So there are two cases:
1. When a time range is selected ( 1w for example ) which is string
2. When user clicks the close icon
for the second one we are getting empty array, so both the situations are being used here.

anwesha-palit-redhat: Did not mean that.
if it is ['1w'] then selected is "1w" and if it is cleared then it is [] or ["0"] or [""] ( either one of this right ? ) so my point was if the type is always String[] then does not this makes it more readable ?
`const selected = Array.isArray(value) && value[0] ? value[0] : null;`

adityavshinde: The type is not always String[], if we select 1 week then it is only string ('1w'), but if it is cleared then it is an empty array.

anwesha-palit-redhat (approved):
/lgtm

PR #1217, opened 2026-07-28
SRVKP-12918: Fix responsive toolbar filter count to show active filter categories instead of total chips

ankrsinha:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Migration
- [ ] CVE Fix

## Summary

Fixes the responsive toolbar badge on PipelineRuns and TaskRuns list pages to display the number of active filter categories instead of the total number of selected filter values.

### Problem:
When the toolbar collapses on smaller viewports, PatternFly counts every chip across all ToolbarFilter components. 
Ex - Selecting both "Cluster" and "Archived" in the Data source filter plus a Time Range counted as 3, even though only 2 filter categories are active.

### Fix:
Added `activeFilterCount` that counts filter categories with at least one selected value, and overrides PF's default count via the `numberOfFiltersText` prop on Toolbar.

## Screen Recordings / Screenshot

https://github.com/user-attachments/assets/7e883e20-d853-4c90-82db-d856011bb1b8


adityavshinde (approved):
Can we change the description's screen recording, as it is showing "1 filters applied" but it is showing correct while I was reviewing ("1 filter applied") ?
The code changes are working fine. 

https://github.com/user-attachments/assets/ed5fb7ca-1c88-454c-98a3-01163c1f1d60

/lgtm

ankrsinha:
> Can we change the description's screen recording, as it is showing "1 filters applied" but it is showing correct while I was reviewing ("1 filter applied") ? The code changes are working fine.
> 
>  Screencast.From.2026-07-28.16-56-44.mp4 
> /lgtm

Okay, i will update the screen recording in the description.

anwesha-palit-redhat (approved):
/lgtm

PR #1227, opened 2026-07-30
SRVKP-12911: icons for PipelineInPipeline(PL) and PipelineInPipelineRuns(PLR)

arvindk-softwaredev:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Migration
- [ ] CVE Fix

## Summary

Adds a new icon next to the PipelineName, if it is a Pipeline in Pipeline in the 
  - PipelineList
  - PipelineRunList
  - Pipeline Details page
  - PipelineRun Details page
  
## Changes

- Updated` PipelineDetailsPage.tsx`, `PipelineRunRow.tsx`, `PipelineRunDetailsPage.tsx`, `PipelineRunsRow.tsx` to display Pipeline in Pipeline icon conditionally
- Added a new method `isPipelineInPipeline` in `pipeline-utils.ts` to check if the PL is a PipelineInPipeline

## Screen Recordings / Screenshot

https://github.com/user-attachments/assets/d36d9e38-4be6-4e25-bee8-36ba7c0fcb1c

https://github.com/user-attachments/assets/ed3f5af6-3dbe-4947-a72d-c95459ac5ec4

### DIFF
--- a/src/components/utils/pipeline-utils.ts
+++ b/src/components/utils/pipeline-utils.ts
@@ -783,13 +783,16 @@ export const getChildPipelinesFromPipelineRun = (
   }));
 };
 
+export const isPipelineInPipeline = (pipeline: PipelineKind): boolean =>
+  pipeline?.spec?.tasks?.some(isPipelineInPipelineTask) ?? false;
+
 export const isPipelineInPipelineTask = (task: PipelineTask): boolean =>
   !!task?.pipelineRef;
 
 export const isPipelineInPipelineRun = (
   pipelineRun: PipelineRunKind,
 ): boolean =>
-  pipelineRun?.status?.childReferences.some(
+  pipelineRun?.status?.childReferences?.some(
     (child) => child.kind === 'PipelineRun',
   ) ||
   pipelineRun?.status?.pipelineSpec?.tasks?.some((item) => item.pipelineRef);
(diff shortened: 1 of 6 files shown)


src/components/utils/pipeline-utils.ts
```diff
@@ -783,13 +783,16 @@ export const getChildPipelinesFromPipelineRun = (
   }));
 };
 
+export const isPipelineInPipeline = (pipeline: PipelineKind): boolean =>
```
anwesha-palit-redhat, line 786: nit: instead of creating a new isPipelineInPipeline can we reuse the existing isPipelineInPipelineTask ? it seems to be checking for the same thing 🤔

arvindk-softwaredev: yes, logic is same just that this `isPipelineInPipeline` will loop through all tasks of a pipeline and do the check.

created this isPipelineInPipeline helper to avoid loops in `PipelineDetailsPage.tsx`, `PipelineRunRow.tsx` for better readability.

wdyt?

anwesha-palit-redhat: makes sense to me, then can we keep both functions but eliminate the duplicated logic by having isPipelineInPipeline delegate to isPipelineInPipelineTask ?

`export const isPipelineInPipelineTask = (task: PipelineTask): boolean =>
  !!task?.pipelineRef;`

`export const isPipelineInPipeline = (pipeline: PipelineKind): boolean =>
  pipeline?.spec?.tasks?.some(isPipelineInPipelineTask) ?? false;`

arvindk-softwaredev: agreed, updated the logic to use `isPipelineInPipelineTask`

anwesha-palit-redhat (approved):
/lgtm

